PR #6, opened 2019-02-12
Missing tests

lkalwa:


spec/neo4j/driver/internal_record_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  subject(:res) do
+    session = driver.session
+    session.run("CREATE (p:Person{names: ['Mark', 'Paul']}) RETURN p.names AS names, ID(p) AS id").single
+  ensure
+    session&.close
+  end
+
+  its(:values) { are_expected.to include(%w[Mark Paul]) }
+  its(:first) { is_expected.to eq(%w[Mark Paul]) }
+  it { expect(res['names']).to eq(%w[Mark Paul]) }
```
klobuczek: What was the reason for testing internal API? The difficulty is that we have 2 implementations which are identical only at the public API. If we want to unit test the internal implementation we would need 2 separate unit tests. Could this be avoided without any downside?

jruby/neo4j/driver/ext/exception_mapper.rb
```diff
@@ -2,7 +2,6 @@
 
 java_import org.neo4j.driver.v1.exceptions.AuthenticationException
 java_import org.neo4j.driver.v1.exceptions.ClientException
-java_import org.neo4j.driver.v1.exceptions.NoSuchRecordException
```
klobuczek: @lkalwa sorry if we already talked about it but why is this exception being removed?

lkalwa: @klobuczek this part is misleading. In exception mapper you're initializing ruby error classes for previously caught `Java::OrgNeo4jDriverV1Exceptions::Neo4jException`, but `NoSuchRecordException` is different type.
See: https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/exceptions/Neo4jException.html
and https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/exceptions/NoSuchRecordException.html.
Not sure if this is the right thing to do, but at least we have less unreachable code.

spec/neo4j/driver/internal_record_spec.rb
lkalwa: Fair point.

jruby/neo4j/driver/ext/exception_mapper.rb
klobuczek: @lkalwa I see, this requires some better handling then. There are few more exceptions we need to handle. I think we need to catch all RuntimeExceptions and `e.reraise` should map the known once and re-raise the others unchanged.

spec/neo4j/driver/internal_record_spec.rb
klobuczek: @lkalwa could we pick up this again?

lkalwa: @klobuczek let's discard these changes.

jruby/neo4j/driver/ext/exception_mapper.rb
lkalwa: @klobuczek I suggest to move this part to another PR, as title of this PR is confusing.

klobuczek:
@lkalwa could clean up this PR to only contain the changes we want to keep?

PR #8, opened 2019-03-28
time_with_zone

klobuczek:


jruby/neo4j/driver/ext/ruby_converter.rb
```diff
@@ -22,13 +22,26 @@ def as_ruby_object
...
+            to_time
           else
             as_object
           end
         end
 
         private
 
+        def to_time
+          time = as_zoned_date_time
```
lkalwa: @klobuczek in general looks good but where is this method from

klobuczek: https://neo4j.com/docs/api/java-driver/current/org/neo4j/driver/v1/Value.html#asZonedDateTime--

lkalwa: I keep forgetting about that magical feature that everything falls back to java

PR #27, opened 2019-06-21
WIP: added zeitwerk gem support

mrhardikjoshi:


lib/neo4j/driver/exceptions/security_exception.rb
```diff
@@ -1,10 +0,0 @@
-# frozen_string_literal: true
-
-module Neo4j
-  module Driver
-    module Exceptions
-      class ProtocolException < Neo4jException
```
mrhardikjoshi: Same class is defined in protocol_exception.rb in same dir [here](https://github.com/neo4jrb/neo4j-ruby-driver/blob/master/lib/neo4j/driver/exceptions/protocol_exception.rb).

I have to remove this one because the file name didn't match the convention of classname and zeitwerk throws the error. Was wondering if this classname was supposed to be SecurityException instead?

klobuczek: this is a mistake it should read `SecurityException`

jruby/neo4j/driver.rb
```diff
@@ -38,5 +13,21 @@ module Net
     module Types
       include_package 'org.neo4j.driver.v1.types'
     end
+
+    def self.after_zeitwerk_load_complete
```
mrhardikjoshi: we needed to execute following code snippet only after zeitwerk load is complete, or else the classes aren't available and unknown constant error is thrown. Hence added this code snippet in `after_zeitwerk_load_complete` to ensure we only call it after the zeitwerk load is complete.

jruby/neo4j/driver/ext.rb
```diff
@@ -0,0 +1,6 @@
+module Neo4j
+  module Driver
```
mrhardikjoshi: A patch/workaround for Jruby version < 9.2.8.0 where the zeitwerk is looking for a file named "jruby/neo4j/driver/ext.rb" to be present and with definition of ext module.

lib/neo4j_ruby_driver.rb
```diff
@@ -2,23 +2,24 @@
...
+require 'zeitwerk'
+require 'zeitwerk/neo4j_ruby_driver_inflector'
+
+loader = Zeitwerk::Loader.new
+loader.tag = "neo4j-ruby-driver"
+lib_dir_path = File.expand_path File.dirname(__FILE__)
+loader.push_dir(lib_dir_path)
+loader.ignore("#{lib_dir_path}/neo4j-ruby-driver_jars.rb")
+
+if ENV['SEABOLT_LIB']&.length&.positive?
```
mrhardikjoshi: cannot shift this in gemspec, so I  have to perform the check for "SEABOLT_LIB" here (only once)

lib/neo4j/driver/exceptions/security_exception.rb
mrhardikjoshi: done, renamed class name to SecurityException, instead of removing the whole file

PR #29, opened 2019-08-06
Improvements

lkalwa:
- cleanups in loading files for MRI and jRuby, 
- corrections in README
- improvements in error handling for missing SEABOLT_LIB for MRI version


.travis.yml
```diff
@@ -24,17 +24,17 @@ env:
   global:
     - JRUBY_OPTS="-J-Xmx1024m -Xcompile.invokedynamic=false -J-XX:+TieredCompilation -J-XX:TieredStopAtLevel=1 -J-noverify -Xcompile.mode=OFF"
     - NEO4J_BOLT_URL="bolt://localhost:7472"
-    - SEABOLT_LIB=./seabolt-1.7.4-Linux-ubuntu-16.04/usr/local/lib/libseabolt17.so
+    - BOLT_LIB=./seabolt-1.7.4-Linux-ubuntu-16.04/usr/local/lib/libseabolt17.so
   matrix:
-    - NEO4J_VERSION=enterprise-3.5.6
-    - NEO4J_VERSION=enterprise-3.4.14
-    - NEO4J_VERSION=community-3.5.6
+    - NEO4J_VERSION=enterprise-3.5.6 SEABOLT_LIB=$BOLT_LIB
```
klobuczek: update neo4j and jruby versions in .travis.yml and README

neo4j-ruby-driver.gemspec
```diff
@@ -33,32 +33,34 @@ Gem::Specification.new do |spec|
...
+  pdir = mri ? 'ffi' : ENV.has_key?('SEABOLT_LIB') ? 'ffi' : 'jruby'
 
   spec.files << Dir['lib/**/*.rb']
   spec.files << Dir["#{pdir}/**/*.rb"]
-  spec.files << Dir["#{pdir}/**/*.jar"]
+
+  unless mri
+    spec.files << Dir["#{pdir}/**/*.jar"]
+    spec.platform = 'java'
+  end
```
klobuczek: the whole above refactoring is incorrect. You are including the jar files even in case of ffi on jruby

neo4j-ruby-driver.gemspec
```diff
@@ -33,32 +33,34 @@ Gem::Specification.new do |spec|
...
-  if ffi
-    spec.add_runtime_dependency 'ffi'
-  else
+  if !mri
     spec.add_runtime_dependency 'jar-dependencies'
     spec.requirements << 'jar org.neo4j.driver, neo4j-java-driver, 1.7.5'
     # avoids to install it on the fly when jar-dependencies needs it
     spec.add_development_dependency 'ruby-maven'
+  else
+    spec.add_runtime_dependency 'ffi'
```
klobuczek: see above

neo4j-ruby-driver.gemspec
```diff
@@ -33,32 +33,34 @@ Gem::Specification.new do |spec|
...
   end
 
   spec.add_runtime_dependency 'activesupport'
-  spec.add_runtime_dependency 'zeitwerk'
+  spec.add_runtime_dependency 'zeitwerk', '= 2.1.6'
```
klobuczek: use the ignore line in loader instead of resorting to older zeitwerk version, see result_stream branch

neo4j-ruby-driver.gemspec
```diff
@@ -33,32 +33,34 @@ Gem::Specification.new do |spec|
     `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   end
 
-  ffi = ENV['SEABOLT_LIB']&.length&.positive?
```
klobuczek: simplify this check everywhere else in the code as well

lib/platform_requirements.rb
```diff
@@ -0,0 +1,10 @@
+#frozen_string_literal: true
+
+# workaround for zeitwerk, without wrapping code in module error:
+# expected file ../../neo4j-ruby-driver/lib/platform_requirements.rb to define constant PlatformRequirements, but didn't
+# is being thrown
```
klobuczek: BTW this is how zeitwerk works. Either the file name matches the module/class name or the file has to be excluded from zeitwerk loading. So this is not a workaround and no explanation is needed.

.travis.yml
lkalwa: I can only update neo4j, ruby is at 9.2.7.0

neo4j-ruby-driver.gemspec
```diff
@@ -33,22 +36,24 @@ Gem::Specification.new do |spec|
...
-  pdir = ffi ? 'ffi' : 'jruby'
+  jruby = RUBY_PLATFORM.match?(/java/)
+  pdir = if ENV.key?('SEABOLT_LIB')
+           'ffi'
+         else
+           jruby ? 'jruby' : 'ffi'
```
klobuczek: this is still incorrect. ffi should be supported on jruby as well in which case the SEABOLT_LIB is required. The last `ffi` is incorrect as this case in supported (not jruby and not seabolt).
Even without understanding anything about the domain any construct like `a ? b : c ? d : b` should always raise a red flag. The conditionals should be built in a way that `b` does not have to be repeated. Following this coding standard helps detecting potential issues.

neo4j-ruby-driver.gemspec
```diff
@@ -33,22 +36,24 @@ Gem::Specification.new do |spec|
...
 
-  spec.platform = 'java' if RUBY_PLATFORM.match?(/java/)
+  spec.add_runtime_dependency 'ffi'
 
-  if ffi
-    spec.add_runtime_dependency 'ffi'
-  else
+  if jruby
+    spec.platform = 'java'
     spec.add_runtime_dependency 'jar-dependencies'
```
klobuczek: this is incorrect as well, we don't want to always include jar dependencies on jruby, ff/jruby does not need them

spec/support/driver_helper.rb
```diff
@@ -13,7 +13,7 @@ def uri
 
   class << self
     def uri
-      ENV.fetch('NEO4J_BOLT_URL', 'bolt://localhost:7687')
+      ENV.fetch('NEO4J_BOLT_URL', 'bolt://localhost:7472')
```
klobuczek: I intentionally changed the port to prevent overwriting the org-api dev database

neo4j-ruby-driver.gemspec
lkalwa: Ok, but in this case if user will install gem with `SEABOLT_LIB` set and then will use it without that variable he will get an error about missing jars

PR #32, opened 2019-09-27
migration to hoe

klobuczek:


spec/spec_helper.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
+$LOAD_PATH.unshift File.join(File.dirname(__FILE__), '..', ENV['SEABOLT_LIB']&.length&.positive? ? 'ffi' : 'jruby')
```
lkalwa: Duplication, but is this load neeeded at all?  Isn't the rspec suite loading Rakefile at the beginning?
I would solve the duplication with ugly Singleton.

klobuczek: yes, I was hoping to find a better solution. I was not able to find a single gem that uses this solution, contrary to what you mentioned.

klobuczek: BTW I don't understand how it works. If I put the LOAD_PATH into Rakefile it's not visible in specs.

PR #43, opened 2020-01-23
Fix for jRuby data type converter

amitsuryavanshi:


spec/neo4j/driver/types/node_spec.rb
```diff
@@ -16,4 +16,7 @@
   its(:labels) { is_expected.to eq(%i[Person]) }
   its(:id) { is_expected.to be_a(Integer) }
   its(:properties) { is_expected.to eq(name: 'John', created: date) }
```
amitsuryavanshi: Not sure how this is passing when `node.properties => {:name=>"John", :created=>#<Java::JavaUtil::Date:0x502f1f4c>}`

PR #54, opened 2020-12-08
Support symbols with value adapter

joshjordan:
We have many, many ad-hoc queries (now using `ActiveGraph::Base.query`) in our application as we upgrade it from `neoj4rb` 9 to `activegraph`. Previously, `Neo4j::ActiveBase.current_session.query` which was our old method of doing ad-hoc queries, did accept symbol params. It looks like this backwards compatability was lost when `activegraph` switched to `neo4j-ruby-driver`.

The majority of these queries use a symbol for at least one of their parameters. `neo4j-ruby-driver` handles a large number of different ruby classes, it seems to me we should handle symbols as well since they are so simple to treat as strings.

### DIFF
--- a/ffi/neo4j/driver/internal/value/value_adapter.rb
+++ b/ffi/neo4j/driver/internal/value/value_adapter.rb
@@ -50,6 +50,9 @@ module Neo4j
               when String
                 object = object.encode(Encoding::UTF_8) unless object.encoding == Encoding::UTF_8
                 Bolt::Value.format_as_string(value, object, object.bytesize)
+              when Symbol
+                object = object.to_s.encode(Encoding::UTF_8)
+                Bolt::Value.format_as_string(value, object, object.bytesize)
               when Hash
                 Bolt::Value.format_as_dictionary(value, object.size)
                 object.each_with_index do |(key, elem), index|
--- a/spec/neo4j/driver/types_spec.rb
+++ b/spec/neo4j/driver/types_spec.rb
@@ -67,6 +67,12 @@ RSpec.describe Neo4j::Driver do
     it { is_expected.to eq param }
   end
 
+  context 'when Symbol' do
+    let(:param) { :symbol }
+
+    it { is_expected.to match(/^symbol$/) }
+  end
+
   context 'when Date in a map' do
     let(:param) { { date: Date.today } }
 


klobuczek:
We have been debating about supporting additional types beyond neo4j types in the driver and decided not to include the support in the driver mainly because this would be a one-way conversion with the type being lost upon storage. The driver is a low level api and we would like to keep it consistent with other official drivers.
In activegraph there are converters between a much richer type system and the neo4j type system. It is possible due to typed properties on ActiveGraph entities. But if you use ActiveGraph to just issue direct cypher queries then the neo4j type system is used (https://neo4j.com/docs/driver-manual/current/cypher-workflow/#driver-type-mapping).

joshjordan:
This passes on MRI but fails on JRuby. The error is non-descript; any additional/obvious details that would help me move this forward? @klobuczek not sure if you're the right person to ask on this repo

joshjordan:
But this is a ruby driver, no? Symbols are probably the most prolific ruby primitve and are widely used due to their performance benefits. Of course, you are right -- we do have to convert them down to strings before going to Cypher, but it seems awfully strange for a ruby driver to not support ruby primitive types as input. Of the dozens of database drivers I've used, I don't think I've ever seen that.

It means that working with the natural ruby types we have all over our system, we have to remember to call `.to_s` on them for every call. We have hundreds of such queries, so it wouldn't make sense to introduce that kind of maintenance overhead - -we'd just end up wrapping the driver with a function that scans for symbols and converts them, I suppose. I would imagine anyone who uses direct queries would do the same. Doesn't that belong in the driver?

joshjordan:
Further, we already do this with hash keys.

image

So, confusingly,

```
filter = { category: 1 }
ActiveGraph::Base.query("MATCH...", filter: filter)
```
is completely valid, but
```
ActiveGraph::Base.query("MATCH...", filterKeys: filter.keys)
```

throws the completely opaque error `no implicit conversion from nil` (actually, using a symbol is currently a stack overflow, but once #53 is merged, it will be this error). It cost me hours just to figure out why my very simple parameters were causing such crazy errors.

klobuczek:
I'm not totally opposed to this automatic one-way conversion of Symbols. Although I am looking at the bigger picture and what makes sense and is consistent with the official neo4j drivers and other database drivers. Using symbols as keys in maps is a different thing than using symbols as data.

What you emphatically state as obvious is not always true:
```
2.7.1 :038 > db.execute "insert into numbers values ( ?, ? )", :three, 3
Traceback (most recent call last):
        1: from (irb):38
RuntimeError (can't prepare Symbol)

```
The above is against sqllite3 driver directly from a new rails app. ActiveRecord does convert Symbols in its attributes to Strings though. So does ActiveGraph.

joshjordan:
Wow, you are right. I’ll admit that I haven’t used the sqlite driver in a
number of years. I have the same reaction about that driver of course,
which is that that doesn’t not provide a good API, but your point is quite
well taken.

Let me ask this: do you expect many folks upgrading from neo4jrb 9 to
active graph? I think it makes sense to rethink your design decisions in a
major upgrade (and I’ll say that many things about active graph and Neo4j
ruby driver are so much more nicely factored than neo4jrb and Neo4j core
was, so big kudos!). But, so far the upgrade path has been pretty painful.

Certainly that’s where my bias comes from in requesting looser constraints
on the API— I had complaints about neo4jrb but the query api was never one
of them 😀

klobuczek:
Yes, let's list the things that could be better designed or expressed in ruby. We have some flexibility, especially around the types. Looks specifically at the temporal types, some of those are ugly hacks. But the major way the driver works is pretty much driven by the official drivers. The same specs must pass on both the neo4j-ruby-driver and neo4j-java-driver (a thin jruby wrapper around the official java driver). We are working now on neo4j-ruby-driver 4.2 and activegraph 11, so it would be a good time to introduce breaking changes.

joshjordan:
:+1: I would be very happy to participate in that discussion. I am going to try to focus on getting my activegraph upgrade done, and then I will come away from that experience with more meanginful discussion points.

klobuczek:
@joshjordan could you refer me to any ruby driver (postgresql, mysql, mongo etc.) that supports symbols as input data?

joshjordan:
Sure thing! Here for the two primary relational db drivers I'm familiar with, which are used by the rails postgres and mysql query adapters respecively:

`pg` gem

image

`mysql2` gem

image

And for mongodb, which has a significantly different query interface and not so simple to just issue a raw query that returns a a param, using the `mongoid` gem, I'm going to use a model query as a proxy:

image

joshjordan:
And for the first two, of course, Rails supports symbols as ActiveRecord parameters, but I figured that wouldn't necessarily prove the drivers do...

klobuczek:
https://rubygems.org/gems/neo4j-ruby-driver/versions/1.7.4
https://rubygems.org/gems/neo4j-java-driver/versions/1.7.3-java

PR #55, opened 2020-12-09
Support non-UTF8 strings

joshjordan:
Another backwards compatibility issue. We insert ASCII 8bit data into String attributes on ActiveNode models. This was supported in `neoj4b` 9 with `neo4j-core`, and is supported by Seabolt, but support for such strings was removed from `neo4j-ruby-driver`in https://github.com/neo4jrb/neo4j-ruby-driver/commit/62c0ce5e33d76de9827643979a8a9fecf6a30477 but that commit does not mention why.

This change appears not to break any tests (at least in my env on MRI -- waiting to see if Travis CI points out any other failures), so if there is some failure case here, we should encode it properly in a test.

I'm not opening this PR because I think its the right change, but rather as a discussion. Should we support ASCII 8 bit strings? If not, why?

### DIFF
--- a/ffi/neo4j/driver/internal/value/value_adapter.rb
+++ b/ffi/neo4j/driver/internal/value/value_adapter.rb
@@ -48,8 +48,12 @@ module Neo4j
               when Types::Bytes
                 Bolt::Value.format_as_bytes(value, object, object.size)
               when String
-                object = object.encode(Encoding::UTF_8) unless object.encoding == Encoding::UTF_8
-                Bolt::Value.format_as_string(value, object, object.bytesize)
+                if object.encoding == Encoding::ASCII_8BIT
+                  Bolt::Value.format_as_bytes(value, object, object.size)
+                else
+                  object = object.encode(Encoding::UTF_8) unless object.encoding == Encoding::UTF_8
+                  Bolt::Value.format_as_string(value, object, object.bytesize)
+                end
               when Hash
                 Bolt::Value.format_as_dictionary(value, object.size)
                 object.each_with_index do |(key, elem), index|


klobuczek:
Thanks for the input. For this one, we may have to question the neo4j type system mapping to ruby type system.
Neo4j has the types String and ByteArray. In Neo4j all Strings are UTF-8 encoded. We mapped those types to ruby String and Neo4j::Driver::Types::Bytes. What you did in this PR is storing the ASCII String as ByteArray in neo4j and would get back Bytes which works seamlessly because Bytes inherits from String. Bytes is merely a marker subclass as String has all the features of ByteArray. But again as with Symbol you write a type into the db but get another type back, which is not perfect.
For your code to work without this PR you would have to change the non-UTF-8 string to Neo4j::Driver::Types::Bytes.new(string).
When we designed the type system mapping we had it reviewed by a few people including neo4j, but I know it is not flawless, so I'm appreciative of any additional feedback even though it might be more difficult to change it now.
Maybe we should remove the Bytes type and have both represented as String and chose the storage type depending on encoding, but that would decrease the number of supported use cases.

joshjordan:
That makes sense, thank you for the explanation. Helps me understand both
the rationale and the underlying technology!

I think this is another case where I'd advocate for having an API that does
not surprise the user. From our perspective, say in the case of Symbol, we
know that Neo4j does not have a symbol type so its not surprising to get a
string back from Neo4j, but it feels unnatural and breaks expectations that
the API does not accept a string. Same for ASCII (or should I say
ByteArray, now that you've explained). From the perspective of the user, we
don't really care how the driver chooses to represent the string. If the
underlying driver and database have support for the string, it is
unexpected and jarring for it to throw an exception when consuming it.

As a side note, I really would like to do something about the error
messages. We don't get an error that says "ASCII strings are not supported,
use a ByteArray instead", we don't get an error that says "No type mapping
available", we actually just get "no implicit conversion of nil", so its
very hard to understand right now. It has taken many hours to both dig into
the code and try modifying my input until I even knew what the cause was...

But back on topic, I think it is the driver's job to provide a reasonable
API that is easy to use. Since we know how to deal with Symbols and ASCII
strings, I think the driver should do so. It could return warnings or even
throw clear exceptions in dev mode (perhaps with a config flag), but right
now it just appears to be broken for certain queries and gives no
indication about why. We considered more than once just giving up on
neo4j-ruby-driver and going back to neo4jrb 9 because there have been so
many backwards compatible issues.

I don't think its realistic to expect the user to call `.to_s` on every
symbol and `Neo4j::Driver::Types::Bytes.new(string)` on every ASCII string
-- that seems to me to be exposing internal implementation details, and
while it has sensible reasoning (thanks again for taking the time to
explain!), in practice I don't think developers are going to do those
things on every call. Like my team, I expect that it just means they are
going to introduce another layer of code that just converts all symbols to
strings and all ASCII strings to ByteArray, which is why I think the driver
should take that on to provide a natural API that behaves as ruby devs
would expect.

Just one person's opinion :)

klobuczek:
Many good points. I'm right now short on time. But one question. Did you get an error when saving ASCII-encoded String? The intended behavior is that string is converted to UTF-8, stored in the database, and on retrieval, you should get UTF-8 encoded string, which granted might not be what you want.

joshjordan:
Yes, in current master it will throw an exception. I think the main
exception is "unable to coerce nil to String", and the embedded exception
is the encoding one: something like "unable to force UTF-8 encoding" and
shows the ASCII character code that failed. This is because the current
master code forces the encoding anytime it is not already UTF-8 and (it
seems) we are working with a string that Ruby cannot convert to UTF-8.

I can check for the specific errors, but yes it currently does. I'm using
my PR's code in my fork to resolve it.

Would it be good for me to write a spec with a failing example? Thanks for
your help!

klobuczek:
So I think we have here a String that is supposedly encoded in ASCII but cannot be converted to UTF-8. That seems to be the origin of the problem. Furthermore, we seem to have an issue with exception handling and that is likely a problem in the driver.
It would be good to have that string, maybe in form of a spec?

joshjordan:
Absolutely. I’ll do that today and I’ll try to illustrate the error to you.
The error handling was a bit hard to follow for me but I’d be happy to take
a pass at addressing it, with some guidance from you.

klobuczek:
So any byte from 128 and higher will cause the conversion problem. I suspect your byte stream is actually not ASCII encoded but as something else e.g. ISO/IEC 8859-1
When storing strings/bytes in neo4j you have to decide:
1. do you want to preserve the text/characters - bytes might be modified
2. do you want to preserve the bytes - bytes are not modified but since the encoding is not stored the text/characters cannot be reproduced anymore without external knowledge. 

In the current implementation, the choice between 1 and 2 is indicated by the class of the object to be stored. This implementation choice is of course disputable, but the clear separation between 1. and 2. is non-negotiable. Furthermore in case 1. the encoding attribute on the String must precisely match the encoding of the byte representation within the String.

I would love to discuss the ruby neo4j type system mapping, but you should first solve the problem at hand not to impede an open-minded discussion.

joshjordan:
Yeah, makes sense. This explanation was helpful, thank you. You are probably right and I should confirm what my byte stream looks like, and I will continue the conversation from there (perhaps with some example specs). Agree that we should solve the problem at hand so we can have a proper discussion about the type system :+1:

klobuczek:
The driver behaves consistently within the implementation choice. The proposed change is not acceptable. Closing the PR.

PR #56, opened 2020-12-10
Support DateTime params

joshjordan:
image

`DateTime` values are currently converted to `Date` values when used as a query parameter because `DateTime` is a subclass of `Date`. Found this in many examples in our codebase where we use `DateTime.current` as a parameter. We could solve this by switching to `Time.now` in our code, but I am trying to future-proof since using a `DateTime` value does NOT throw an error nor does the query log show anything wrong (because it shows the string value of the ruby object). The only reason I caught this was because one of our ActiveNode models was throwing an error that it could not convert a `Date` object to `DateTime` when reading a `property`, but I had written a `DateTime` to the field and went digging through the code.

### DIFF
--- a/ffi/neo4j/driver/internal/value/value_adapter.rb
+++ b/ffi/neo4j/driver/internal/value/value_adapter.rb
@@ -63,6 +63,8 @@ module Neo4j
                 object = object.to_a
                 Bolt::Value.format_as_list(value, object.size)
                 object.each_with_index { |elem, index| to_neo(Bolt::List.value(value, index), elem) }
+              when DateTime
+                TimeWithZoneOffsetValue.to_neo(value, object)
               when Date
                 DateValue.to_neo(value, object)
               when ActiveSupport::Duration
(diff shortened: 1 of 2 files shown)


klobuczek:
@joshjordan fixed in https://github.com/neo4jrb/neo4j-ruby-driver/commit/86529ab1f60accf218dea16d05f59a668842ab5c
Any pull request should make sure that bothand  ffi and jruby flavors pass the specs. Unfortunately, travis stopped building our projects for free, so that does not help.

klobuczek:
https://rubygems.org/gems/neo4j-ruby-driver/versions/1.7.4
https://rubygems.org/gems/neo4j-java-driver/versions/1.7.3-java

joshjordan:
Thanks for that! I don’t have the Java side running on my machine so I’m
only able to test against MRI right now. Was relying on Travis to prove or
disprove my work. Appreciate the release

PR #60, opened 2021-03-10
Pass statement bytesize (not char length) to set_run_cypher

### LINKED ISSUE #59: Multi-byte utf-8 characters cause queries to be truncated
I've encountered a bug with queries that include multi-byte utf-8 characters. This code produces an exception due to the query that is sent to neo4j being truncated two characters early:

```
Neo4j::Driver::GraphDatabase.driver('bolt://localhost:7687',
                                    Neo4j::Driver::AuthTokens.basic('REDACTED', 'REDACTED'),
                                    encryption: false) do |driver|
  driver.session do |session|
    session.read_transaction do |tx|
      tx.run("MATCH (t:Term{ uri: 'http://eol.org/schema/terms/Löbbecke_Museum_Düsseldorf' })-[:synonym_of]->(parent:Term) RETURN t LIMIT 1")
    end

    puts 'success'
  end
end
```
```
neo4j-ruby-driver-1.7.4/ffi/neo4j/driver/internal/handlers/response_handler.rb:38:in `check_summary_failure': Unexpected end of input: expected whitespace or an expression (line 1, column 124 (offset: 123)) (Neo4j::Driver::Exceptions::ClientException)
"MATCH (t:Term{ uri: 'http://eol.org/schema/terms/Löbbecke_Museum_Düsseldorf' })-[:synonym_of]->(parent:Term) RETURN t LIMIT"
```

I believe this is due to the character length -- and not byte length -- of the query string being passed to the seabolt `set_run_cypher` method. If I pad the original query with two or more trailing spaces, it succeeds. I've tested the change to  use byte length, and the error goes away. I'm preparing a pull request now.

mvitale:
Fixes #59

### DIFF
--- a/ffi/neo4j/driver/internal/async/direct_connection.rb
+++ b/ffi/neo4j/driver/internal/async/direct_connection.rb
@@ -22,7 +22,7 @@ module Neo4j
 
           def write_and_flush(statement, parameters, boomarks_holder, config, run_handler, pull_handler)
             check_error Bolt::Connection.clear_run(bolt_connection)
-            check_error Bolt::Connection.set_run_cypher(bolt_connection, statement, statement.size, parameters.size)
+            check_error Bolt::Connection.set_run_cypher(bolt_connection, statement, statement.bytesize, parameters.size)
             parameters.each_with_index do |(name, object), index|
               name = name.to_s
               Value::ValueAdapter.to_neo(


klobuczek (approved):

mvitale:
@klobuczek yep, fair enough. We usually do that :) Thanks for merging.

PR #65, opened 2021-09-02
WIP

mrhardikjoshi:
1) made AuthTokens consistent in all specs

### DIFF
--- a/spec/neo4j/driver/dev_manual_examples_spec.rb
+++ b/spec/neo4j/driver/dev_manual_examples_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe Neo4j::Driver do
     after { driver2.close }
 
     let(:driver2) { Neo4j::Driver::GraphDatabase.driver(uri, auth_tokens, config) }
-    let(:auth_tokens) { Neo4j::Driver::AuthTokens.basic('neo4j', 'pass') }
+    let(:auth_tokens) { basic_auth_token }
     let(:config) { {} }
 
     context 'Example 2.1. The driver lifecycle' do
--- a/spec/integration/session_spec.rb
+++ b/spec/integration/session_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe 'SessionSpec' do
   end
 
   it 'handles nil config' do
-    driver = Neo4j::Driver::GraphDatabase.driver(uri, Neo4j::Driver::AuthTokens.basic('neo4j', 'pass'))
+    driver = Neo4j::Driver::GraphDatabase.driver(uri, basic_auth_token)
     session = driver.session
     session.close
     expect(session).not_to be_open
--- a/spec/neo4j/driver/simplified_examples_spec.rb
+++ b/spec/neo4j/driver/simplified_examples_spec.rb
@@ -35,9 +35,7 @@ RSpec.describe Neo4j::Driver do
   end
 
   it 'Driver with block and fetching before session close' do
-    username = 'neo4j'
-    password = 'pass'
-    result = Neo4j::Driver::GraphDatabase.driver(uri, Neo4j::Driver::AuthTokens.basic(username, password)) do |driver|
+    result = Neo4j::Driver::GraphDatabase.driver(uri, basic_auth_token) do |driver|
       driver.session { |session| session.run('CREATE (a:Person {name: $name}) RETURN a.name', name: 'John').single }
     end
 
--- a/spec/support/cluster_extension.rb
+++ b/spec/support/cluster_extension.rb
@@ -4,8 +4,6 @@ require 'neo4j/driver/util/cc/cluster_control'
 require 'neo4j/driver/util/cc/shared_cluster'
 
 RSpec.shared_context 'cluster_extension' do
-  USER = 'neo4j'
-  PASSWORD = 'pass'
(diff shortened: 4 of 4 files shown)


spec/neo4j/driver/dev_manual_examples_spec.rb
```diff
@@ -23,7 +23,7 @@
     after { driver2.close }
 
     let(:driver2) { Neo4j::Driver::GraphDatabase.driver(uri, auth_tokens, config) }
-    let(:auth_tokens) { Neo4j::Driver::AuthTokens.basic('neo4j', 'pass') }
+    let(:auth_tokens) { basic_auth_token }
```
klobuczek, line 26: maybe make here an exception as this is a documentation spec taken from Neo4j Driver Manual and there is merit in showing how this works without chasing custom helper methods.

ruby/neo4j/driver/ext/config_converter.rb
```diff
@@ -31,6 +31,8 @@ def config_method(key, value)
           when 'resolver'
             proc = value
             value = ->(address) { java.util.HashSet.new(proc.call(address)) }
+          when 'trust_strategy'
+            value = Java::OrgNeo4jDriver::Config::TrustStrategy.trust_all_certificates
```
klobuczek, line 35: so this does not match the above. If you replace a class like TrustStrategy you have to be able to replace all its usages across the entire code base. That's why it is not a good idea to start with simple terminal (leaf) classes but rather with classes that are only used from already converted ruby code. There is a bit of those and with each conversion more are added to this set.
In other words, the ruby code and the still-active java code together should at no point reference both ruby and java version of a the same class.

ruby/neo4j/driver/internal/scheme.rb
```diff
@@ -5,28 +5,29 @@ module Neo4j::Driver::Internal::Scheme
   NEO4J_URI_SCHEME = 'neo4j'
   NEO4J_HIGH_TRUST_URI_SCHEME = 'neo4j+s'
   NEO4J_LOW_TRUST_URI_SCHEME = 'neo4j+ssc'
-
-  def validate_scheme!(scheme)
-    unless [BOLT_URI_SCHEME, BOLT_LOW_TRUST_URI_SCHEME, BOLT_HIGH_TRUST_URI_SCHEME, NEO4J_URI_SCHEME,
-            NEO4J_LOW_TRUST_URI_SCHEME, NEO4J_HIGH_TRUST_URI_SCHEME].include?(scheme)
-      raise ArgumentError, scheme ? "Invalid address format #{scheme}" : 'Scheme must not be null'
+  class << self
```
klobuczek, line 8: The Schema class in java has only static methods, so it can be represented in ruby by a module as no instances are created. Did you have difficulties calling the `validate_scheme!` method the way it was set up?

ruby/neo4j/driver/ext/config_converter.rb
mrhardikjoshi: Yup, I agree, making changes to not leave out same class in both ruby-java 👍

ruby/neo4j/driver/internal/scheme.rb
mrhardikjoshi: The reason I made it module level method is because I don't want to include `Scheme` in `DriverFactory` class.

Out of all code in `Scheme` module (constants and bunch of methods) only a single method was being used there. also  in `GraphDatabase` class another single method of `Scheme` was being used. I think it is better if we keep `Scheme` as a standalone module, and use it directly instead of calling it. Let me know if you think ottherwise.

ruby/neo4j/driver/ext/config_converter.rb
klobuczek, line 35: the value has to be taken from the `value` parameter rather than hardcoded.

ruby/neo4j/driver/config.rb
```diff
@@ -4,12 +4,15 @@ module Neo4j
...
-      TrustStrategy = Java::OrgNeo4jDriver::Config::TrustStrategy
+      class TrustStrategy
+        TRUST_ALL_CERTIFICATES = :trust_all_certificates
+        TRUST_CUSTOM_CA_SIGNED_CERTIFICATES = :trust_custom_ca_signed_certificates
+        TRUST_SYSTEM_CA_SIGNED_CERTIFICATES = :trust_system_ca_signed_certificates
+
+        def self.trust_all_certificates
+          TRUST_ALL_CERTIFICATES
+        end
+      end
```
klobuczek, line 28: this class is incomplete. I would not replace the java class if you don't have a complete ruby replacement.

klobuczek:
In general, I think you have started with an apparently simple class but with many details and marginal functionality.

ruby/neo4j/driver/config.rb
mrhardikjoshi: @klobuczek yes, I am completing it along with it's dependent classes.

mrhardikjoshi:
@klobuczek please review this PR.
There is only one spec failing (spec/integration/causal_clustering_spec.rb:173 - undefined method `routing_table' for nil:NilClass) which is also failing in master.

There are few more changes, but I am creating another PR for them to keep review and changes easy manageable.

PR #66, opened 2021-09-13
Updated github workflow to have ruby 3.0.2 build

mrhardikjoshi:
- [x] updated github workflow to have ruby 3.0.2 build
- [x] Update driver to not include java classes when not using jruby

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,12 +12,18 @@ on:
 jobs:
   test:
     runs-on: ubuntu-18.04
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


mrhardikjoshi:
@klobuczek added a build to run against ruby 3.0.2 in this PR, It is currently failing for all test because driver loader and config files have java classes mentioned in them. It should start passing once we remove those java classes references.

Let me know if you think I should merge this in #65 instead of master

klobuczek:
@mrhardikjoshi at this point there is no benefit of including MRI build as the CI will always fail but we want to be in a position to accept PRs and deploy patch releases, so the build always failing is not good. I do believe we should run this type of build manually.

PR #70, opened 2021-10-10
add info about recent neo4j versions

slicksammy:
I spent a lot of time today trying to set this up and turns out it won't work with the neo4j version I'm running (4.3.5)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,9 @@ Or install it yourself as:
 
     $ gem install neo4j-ruby-driver
 
-Note: Seabolt has not been maintained for a while so the ruby driver will not work with more recent Neo4j versions (https://github.com/neo4jrb/neo4j-ruby-driver/issues/63#issuecomment-905394782)
+## Server Compatibility
+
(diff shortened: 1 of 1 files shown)


klobuczek:
The driver compatibility matrix is located at https://support.neo4j.com/hc/en-us/articles/115013134648-Neo4j-Supported-Versions
None of the official drivers duplicate this information in their own documentation.
Note: neo4j-ruby-driver 1.7 works as well with neo4j 4.0 in the context of activegraph but it is probably not fully compatible.

slicksammy:
This is the main page on their site https://neo4j.com/developer/ruby/ and it does not link to the matrix. I don't understand the down side of adding this snippet. Imo even if it saves one person one hour it's still worth it.

slicksammy:
Another idea is to put a link to the matrix instead

slicksammy:
@klobuczek what do you think? can we add the matrix to the readme?

klobuczek (approved):

PR #71, opened 2021-10-19
Support for trust strategy

mrhardikjoshi:
1) made AuthTokens consistent in all specs
2) Migration of `org.neo4j.driver.internal.SecuritySettings`, `Config.TrustStrategy` and `org.neo4j.driver.internal.RevocationStrategy`
3) removed few dependency of `Config` class of java (being dependent by use of `java_config` method). Dependency pending for java-Config is 1) `routing_settings` 2) `eventLoopThreads` and 3) `logging` and few other places.


ruby/neo4j/driver/internal/security_setting.rb
```diff
@@ -0,0 +1,67 @@
...
+      else
+        org.neo4j.driver.internal.security.SecurityPlanImpl.forAllCertificates(false, RevocationStrategy.NO_CHECKS)
+      end
+    end
+
+    private
+
+    def assert_security_settings_not_user_configured(uri_scheme)
+      if customized?
+        raise Neo4j::Driver::Exceptions::ClientException.new("Scheme #{uri_scheme} is not configurable with manual encryption and trust settings")
```
klobuczek, line 41: java passes the cause to the client exception, we have something similar in ruby Neo4jException, but I think it does not accept 2 params (cause and exception), we might have to change the initialize method to use named params

PR #82, opened 2021-12-02
Security plan impl conversion

Kinju011:
converted security_plan_impl from java to ruby


ruby/neo4j/driver/internal/security/security_plan_impl.rb
```diff
@@ -3,10 +3,90 @@ module Security
...
+
+          trust_manager_factory = javax.net.ssl.TrustManager.get_instance(javax.net.ssl.TrustManagerFactory.get_default_algorithm)
+          trust_manager_factory.init(javax.net.ssl.CertPathTrustManagerParameters.new(pkix_builder_parameters))
+          ssl_context.init(javax.net.ssl.KeyManager[0].new, trust_manager_factory.get_trust_managers, nil)
+          ssl_context
+        end
+
+        def load_system_certificates(trusted_key_store)
+          temp_factory = javax.net.ssl.TrustManagerFactory.get_instance(javax.net.ssl.TrustManagerFactory.get_default_algorithm)
+          temp_factory.init(java.security.KeyStore.java_class.cast(nil))
```
klobuczek, line 58:
```suggestion
          temp_factory.init(nil)
```
We assume all code will be in pure ruby, so we will have to replace all the java libraries. But at this point, we don't go that far.

ruby/neo4j/driver/internal/security/security_plan_impl.rb
```diff
@@ -3,10 +3,90 @@ module Security
...
+            if trust_manager.kind_of?(javax.net.ssl.X509TrustManager)
+              x509_trust_manager = javax.net.ssl.X509TrustManager.java_class.cast(trust_manager)
+              break
+            end
+          end
+
+          if x509_trust_manager.nil?
+            raise Neo4j::Driver::Exceptions::CertificateException, 'No system certificates found'
+          else
+            org.neo4j.driver.internal.util.CertificateTool.load_x509_cert(x509_trust_manager.get_accepted_issuer, trusted_key_store)
```
klobuczek, line 71: replace with ruby class and anywhere else where I did not comment

klobuczek (changes requested):
This is great. Let's just zero in on the mentioned aspects.

klobuczek (approved):

ruby/neo4j/driver/internal/security/security_plan_impl.rb
```diff
@@ -3,10 +3,90 @@ module Security
...
+
+          if x509_trust_manager.nil?
+            raise Neo4j::Driver::Exceptions::CertificateException, 'No system certificates found'
+          else
+            Util::CertificateTool.load_x509_cert(x509_trust_manager.get_accepted_issuer, trusted_key_store)
+          end
+        end
+      end
+
+      class TrustAllTrustManager < javax.net.ssl.X509TrustManager
```
klobuczek, line 76: `X509TrustManager` is an interface. Interfaces without a default implementation we should simply skip. Otherwise, the class cannot be loaded and the specs won't pass even if the classes are not used.

klobuczek (changes requested):
Let's see if we can have those classes successfully loaded. Otherwise we will have to put them all in a separate directory which zeitwerk does not try to load.

klobuczek (approved):

PR #88, opened 2021-12-10
Internal async conversion

Kinju011:


ruby/neo4j/driver/internal/async/internal_async_transaction.rb
```diff
@@ -0,0 +1,18 @@
...
+
+        def initialize(tx)
+          @tx = tx
+        end
+
+        delegate :commit_async, :rollback_async, :open?, to: :@tx
+
+        def run_async(query)
+          @tx.run_async(query)
+        end
```
klobuczek, line 14: the delegation works even methods with parameters

ruby/neo4j/driver/internal/async/internal_async_transaction.rb
```diff
@@ -0,0 +1,18 @@
+module Neo4j::Driver
+  module Internal
+    module Async
+      class InternalAsyncTransaction
```
klobuczek, line 4: missing base class `AsyncAbstractQueryRunner`

PR #98, opened 2021-12-20
reactive package migration

mrhardikjoshi:


ruby/neo4j/driver/internal/util/error_util.rb
```diff
@@ -0,0 +1,113 @@
...
+
+        exception_class.new(code, message)
+      end
+
+      def fatal?(error)
+        return true unless error.is_a?(Neo4j::Driver::Exceptions::Neo4jException)
+
+        error_code = error.code&.downcase
+        return true if protocol_violation_error?(error_code)
+        return false if client_or_transient_error?(error_code)
```
mrhardikjoshi, line 50: @klobuczek what do you think of this comment?

should I change this code like below? It will be different than original but more accurate as per ruby standard.

```suggestion
        protocol_violation_error?(error_code)
```

klobuczek: neither the implementation not the suggestion is correct

ruby/neo4j/driver/internal/util/error_util.rb
```diff
@@ -0,0 +1,113 @@
...
+                            Neo4j::Driver::Exceptions::DatabaseException
+                          end
+
+        exception_class.new(code, message)
+      end
+
+      def fatal?(error)
+        return true unless error.is_a?(Neo4j::Driver::Exceptions::Neo4jException)
+
+        error_code = error.code&.downcase
```
klobuczek, line 48: why downcasing?

ruby/neo4j/driver/internal/util/error_util.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      def extract_class_from_code(code, parts_counts)
+        parts = code.split('.')
+        parts.length < parts_counts ? '' : parts[parts_counts - 1]
+      end
+
+      def protocol_violation_error?(error_code)
+        return false if error_code.nil?
+
+        error_code.start_with?('neo.clienterror.request')
```
klobuczek, line 103: That's overinterpreting. The error code comes from the server and is well defined.

ruby/neo4j/driver/internal/util/error_util.rb
```diff
@@ -0,0 +1,113 @@
...
+      end
+
+      def client_or_transient_error(error_code)
+        return false if error_code.nil?
+
+        error_code.include?('clienterror') || error_code.include?('transienterror')
```
klobuczek, line 109: keep the case

ruby/neo4j/driver/access_mode.rb
```diff
@@ -0,0 +1,16 @@
...
+# This classification is key for routing driver to route transactions to a cluster correctly.
+#
+# While any {@link AccessMode} will be ignored while running transactions via a driver towards a single server.
+# As the single server serves both read and write operations at the same time.
+module Neo4j::Driver::AccessMode
+  # Use this for transactions that requires a read server in a cluster
+  READ = Java::OrgNeo4jDriver::AccessMode::READ # TO DO: replace this with :read once all java refs are done
+
+  # Use this for transactions that requires a write server in a cluster
+  WRITE = Java::OrgNeo4jDriver::AccessMode::WRITE # :write # TO DO: replace this with :write once all java refs are done
```
klobuczek, line 15: to take it further we decided with @Kinju011 to convert all dependent classes. The code needs to load successfully but not necessarily execute. Such a conversion will help us get rid of any java sources and only depend on netty and reactive jars, although I did not plan to release reactive at first release at all, now that you converted it we can keep it. But let's don't have any dependency on `Java::OrgNeo4jDriver` in converted code.

ruby/neo4j/driver/internal/reactive/abstract_rx_query_runner.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Neo4j::Driver::Internal::Reactive
+  module AbstractRxQueryRunner
+    def parse_query(query, opts)
+      return if query.is_a?(Neo4j::Driver::Query)
```
klobuczek, line 6: let's don't implement method overloading.

ruby/neo4j/driver/internal/reactive/internal_rx_result.rb
```diff
@@ -0,0 +1,88 @@
...
+    IGNORE = org.neo4j.driver.internal.shaded.reactor.core.publisher.FluxSink::OverflowStrategy::IGNORE
+
+    def initialize(&cursor_future)
+      @cursor_future_supplier = cursor_future
+    end
+
+    def keys
+      org.neo4j.driver.internal.shaded.reactor.core.publisher.Mono.defer do
+        org.neo4j.driver.internal.shaded.reactor.core.publisher.Mono.from_completion_stage(cursor_future)
+           .map(&Neo4j::Driver::Internal::Cursor::RxResultCursor.method(:keys))
```
klobuczek, line 14:
```suggestion
           .map(&Cursor::RxResultCursor.method(:keys))
```

ruby/neo4j/driver/internal/util/error_util.rb
```diff
@@ -0,0 +1,113 @@
...
+module Neo4j::Driver::Internal::Util
+  class ErrorUtil
+    DEFAULT_CONN_TERMINATED_REASON = 'Please ensure that your database is listening on the correct host and port and '\
+      'that you have compatible encryption settings both on Neo4j server and driver. '\
+      'Note that the default encryption setting has changed in Neo4j 4.0.'.freeze
+
+    SEC_EXCEPTION_CODE_MAPPING = {
+      'neo.clienterror.security.unauthorized': Neo4j::Driver::Exceptions::AuthenticationException,
+      'neo.clienterror.security.authorizationexpired': Neo4j::Driver::Exceptions::AuthorizationExpiredException,
+      'neo.clienterror.security.tokenexpired': Neo4j::Driver::Exceptions::TokenExpiredException
```
klobuczek, line 12: do not downcase

klobuczek (changes requested):

ruby/neo4j/driver/access_mode.rb
klobuczek: we decided as well to keep java dependencies in package notation rather than ruby notation to easily spot the places

ruby/neo4j/driver/internal/util/error_util.rb
mrhardikjoshi: I see the use of `equalsIgnoreCase`, this gave me the impression that I shouldn't rely on error code's case. hence the down casing and comparing.
https://github.com/neo4j/neo4j-java-driver/blob/5.0/driver/src/main/java/org/neo4j/driver/internal/util/ErrorUtil.java#L74

ruby/neo4j/driver/internal/reactive/abstract_rx_query_runner.rb
mrhardikjoshi: hmm ok, shall I remove this module `AbstractRxQueryRunner`?

ruby/neo4j/driver/access_mode.rb
mrhardikjoshi: There are couple of places in code and in test where `Neo4j::Driver::AccessMode::READ/WRITE` are used ([example](https://github.com/neo4jrb/neo4j-ruby-driver/blob/ec429228d3eb3efea4d2c25af7f6c791d9cbb3e1/spec/neo4j/driver/dev_manual_examples_spec.rb#L122)). 

those places were passing java enum in java function call making the test pass. Since I introduced this module that code started breaking as it is now passing ruby symbol instead java enum.

I guess I should remove this module and not migrate it till we have removed such dependencies.

ruby/neo4j/driver/internal/util/error_util.rb
mrhardikjoshi: now sure what you are referring to.

ruby/neo4j/driver/access_mode.rb
klobuczek: I didn't realize that we already have Neo4j::Driver::AccessMode through https://github.com/neo4jrb/neo4j-ruby-driver/blob/f647af00dad736602e2622feb25f689ef71f5957/ruby/neo4j/driver.rb#L18
All guess those `include_package` will cause more problems. I'll try to tackle those conflicts.

ruby/neo4j/driver/internal/reactive/abstract_rx_query_runner.rb
klobuczek: I think so.

ruby/neo4j/driver/internal/util/error_util.rb
klobuczek: both your 1st version and your suggested change do not behave like the java code, not even if you assume that nil is equivalent to false.

ruby/neo4j/driver/internal/util/error_util.rb
klobuczek: Yes, that's surprising. I think java code is inconsistent on that. In any case, do not change literals.

ruby/neo4j/driver/access_mode.rb
klobuczek: Maybe rather than removing the module, convert it to its final state and comment out the conflicting constants.

mrhardikjoshi: when I keep this module with constants commented I get `uninitialized constant Neo4j::Driver::AccessMode::READ`. so removing it.

klobuczek (approved):

PR #103, opened 2021-12-25
internal files migrated

Kinju011:


ruby/neo4j/driver/internal/bolt_server_address.rb
```diff
@@ -0,0 +1,85 @@
+module Neo4j::Driver
+  module Internal
+    # Holds a host and port pair that denotes a Bolt server address.
+    class BoltServerAddress < Struct.new(:host, :connection_host, :port, :string_value)
+      extend Net::ServerAddress
```
Kinju011, line 5: @klobuczek 
Here this line causes the issue of failing specs, I have tried it with include as well but got the same result.
when I remove this line then all specs are passing
so, need your suggestions on this.
or should I define a method of a module in this class?

also, I have tried to remove 
```
module Net
  include_package 'org.neo4j.driver.net'
end
```
from "driver.rb" file. but it also fails the specs

klobuczek: The problem is deeper than that. It is impossible to leverage the BoltServerAddress until all java code creating instances of the class is converted.
I've put the 2 classes on the sidetrack. The next attempt we can do when ResolvedBoltServerAddress and cluster/RediscoveryImpl are converted.

klobuczek (approved):

PR #104, opened 2021-12-27
driver summary migrated

Kinju011:


ruby/neo4j/driver/summary/query_type.rb
```diff
@@ -0,0 +1,12 @@
+module Neo4j
+  module Driver
+    module Summary
```
mrhardikjoshi, line 3: I think the specs are failing because this namespace is being used here https://github.com/neo4jrb/neo4j-ruby-driver/blob/master/ruby/neo4j/driver.rb#L28-L30

Kinju011: Okay, will look into it.

klobuczek: Well, the class is not completely converted.  Same problem as with the last PR. Once MetadataExtractor usages are converted we can move ahead. I think this PR should contain the conversion of all internal/handlers. This is the highest priority now.

klobuczek: One more thing:
`include_package` does not redefine an already loaded constant. So if you convert a class, the converted class will be used from ruby, however, unconverted java code will still use the java class. That's why we need to convert all parents in the `usage` graph. I'd be ok, with converting everything without passing specs.
I currently don't have enough ruby code to start building the blocks together.

PR #123, opened 2022-02-21
Fix DriverFactory and PullAllResponseHandler to be able to connect to Neo4j Aura

krzyczak:
Without these changes there are two issues.

Issue 1: The driver cannot connect to the neo4j instance
The `Neo4j::Driver::Exceptions::ClientException: Unsupported URI scheme: neo4j+s` error is thrown.
Fix: e10de99cb29c417f00a0e5e7134110bf3585c423

Issue 2: Empty `:labelsOrTypes`
Sometimes when querying (e.g. loading schema using the activegraph gem), when index records are accessed the value of `:labelsOrTypes` is an empty array (`[]`), so the code throws an `NoMethodError: undefined method `to_sym' for nil:NilClass`. This error is coming for example from here https://github.com/neo4jrb/activegraph/blob/master/lib/active_graph/core/schema.rb#L64
Fix: d71109888d9f32f7b9809abba0f2afd049c1456a

### DIFF
--- a/ffi/neo4j/driver/internal/driver_factory.rb
+++ b/ffi/neo4j/driver/internal/driver_factory.rb
@@ -8,6 +8,7 @@ module Neo4j
         BOLT_URI_SCHEME = 'bolt'
         BOLT_ROUTING_URI_SCHEME = 'bolt+routing'
         NEO4J_URI_SCHEME = 'neo4j'
+        NEO4J_AURA_URI_SCHEME = 'neo4j+s'
         DEFAULT_PORT = 7687
 
         def new_instance(uri, auth_token, config)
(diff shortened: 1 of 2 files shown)


klobuczek:
@krzyczak 
Issue1: 
The driver 1.7.x does not support the +s and +ssc schema variants. They could be potentially implemented but it's not worth it as the same effect can be achieved with configuration. Furthermore, official drivers 1.7 do not support those variants either. Our 4.4 jruby driver does support them and so will the upcoming pury ruby driver.
Aura does not see the schema, it is only of use for the driver. The way you implemented it is equivalent to passing bolt://host:port. The ffi directory is obsolete on master it will be deleted with the release of 4.4 pure ruby driver. Any PRs towards 1.7 should be made against the 1.7 branch.

Issue2:
This is an activegraph issue and should be solved over there. The driver is completely agnostic of db schema. The schema of index description has been changing in neo4j from version to version. https://github.com/neo4jrb/activegraph/blob/master/lib/active_graph/core/schema.rb attempts to deal with all the different variants. Apparently still somewhat unsuccessfully. So there might be a need for improvement.

Reminder: 
Driver 1.7 fully supports neo4j version 3.5 only with limited support for 4.0 (to the extent needed for activegraph).

krzyczak:
Than you for an advice. I will close this PR now in favour of the ActiveGraph update https://github.com/neo4jrb/activegraph/pull/1668

PR #142, opened 2022-05-05
resolver_resolution request implemented

Kinju011:


testkit-backend/lib/testkit/backend/messages/requests/resolver_resolution_completed.rb
```diff
@@ -2,6 +2,11 @@ module Testkit::Backend::Messages
   module Requests
     class ResolverResolutionCompleted < Request
       def process
+        named_entity('ResolverResolutionRequired', id: requestId, address: address)
```
Kinju011, line 5: here I am not sure how to get address 
I have checked `NewDriver`'s `callback_resolver` but need your suggestion on it.

klobuczek: this is not the correct response. When testkit-backend responds to NewDriver with ResolverResolutionRequired then testkit responds with the request ResolverResolutionCompleted and then testkit-backend responds with Driver.

PR #144, opened 2022-05-05
changes in transaction commit request and response

Kinju011:

PR #145, opened 2022-05-05
transaction rollback response changed

Kinju011:


klobuczek:
see https://github.com/neo4jrb/neo4j-ruby-driver/pull/144

Kinju011:
> see #144

okay

PR #146, opened 2022-05-06
updated forced routing table request

Kinju011:


testkit-backend/lib/testkit/backend/messages/requests/forced_routing_table_update.rb
```diff
@@ -5,8 +5,12 @@ def process
...
       end
 
-      # def to_object
-      # end
+      def to_object
+        fetch(driverId).session(
+          database: database,
+          bookmarks: bookmarks&.map { |bookmark| Neo4j::Driver::Bookmark.from(Array(bookmark)) }
+          )
+      end
```
klobuczek, line 13: in the current form the code is unreachable.

klobuczek: java testkit-backend does not implement this request so we don't know how to implement it. I'm not sure if the routing table is updated just by creating a session.

klobuczek (changes requested):

testkit-backend/lib/testkit/backend/messages/requests/forced_routing_table_update.rb
Kinju011: okay

klobuczek:
java and python testkit-backends do not implement this request

PR #150, opened 2022-05-09
driver_metrics config added

Kinju011:


klobuczek:
@Kinju011 in Neo4j::Driver::Ext::ConfigConverter handle `driver_metrics` in similar way `encryption` is handled. Avoid any duplication.

Kinju011:
> @Kinju011 in Neo4j::Driver::Ext::ConfigConverter handle `driver_metrics` in similar way `encryption` is handled. Avoid any duplication.

okay, thanks
will do that.

jruby/neo4j/driver/ext/config_converter.rb
```diff
@@ -19,6 +19,9 @@ def config_method(key, value)
           when 'encryption'
             method = :without_encryption unless value
             value = nil
+          when 'driver_metrics'
+            method = :without_driver_metrics unless value
+            value = nil
```
klobuczek: as I mentioned earlier this should be solved in a more DRY way

klobuczek (changes requested):

jruby/neo4j/driver/ext/config_converter.rb
Kinju011: okay, updating it it.

klobuczek (approved):

PR #173, opened 2022-06-21
increased timeout minutes for github actions

mrhardikjoshi:
increased timeout from 10 to 20 as for ruby builds it was taking more time resulting in github canceling the jobs.


mrhardikjoshi:
Increased timeout doesn't work as some test are running for indefinite time. Hence we have to fix them first to make all builds work

PR #176, opened 2022-07-01
concurrent promise future

Kinju011:


klobuczek:
I was going to suggest merging it the way it is. We will still have to do some work once the AutoPulllRequestHandler is completely fixed. However we have a TLS spec failing on jruby. What changed?

Kinju011:
> I was going to suggest merging it the way it is. We will still have to do some work once the AutoPulllRequestHandler is completely fixed. However we have a TLS spec failing on jruby. What changed?

yes, I have noticed that and checking it.

klobuczek (approved):

PR #183, opened 2022-07-11
failing specs fixes

Kinju011:


ruby/neo4j/driver/internal/handlers/legacy_pull_all_response_handler.rb
```diff
@@ -65,7 +64,7 @@ def disable_auto_read_management
         end
 
         def peek_async
-          record = @records.first
+          record = @records.items.first
```
klobuczek: this does not work, have a look at AutoPull...
this method needs to wait under some circumstances

klobuczek (changes requested):

ruby/neo4j/driver/internal/handlers/legacy_pull_all_response_handler.rb
Kinju011: okay

ruby/neo4j/driver/internal/handlers/legacy_pull_all_response_handler.rb
```diff
@@ -65,7 +64,10 @@ def disable_auto_read_management
         end
 
         def peek_async
-          record = @records.first
+          while @records.empty? && !@record_future.nil?
```
klobuczek: in general every method you touch should not have the word `future` in it. Most of the time you can just remove the future or replace it with `@record`. He we don't need the extra `@record` variable as we work with wait/signal and `@records`. The done condition here is  `@ignore_records || @finished`. It was already there in the code.

klobuczek (changes requested):

klobuczek (approved):

PR #198, opened 2022-09-14
fixed the example spec with neo4j scheme

Kinju011:


lib/neo4j/driver/exceptions/socket_exception.rb
```diff
@@ -0,0 +1,11 @@
...
+
+module Neo4j
+  module Driver
+    module Exceptions
+      # While resolving an address using getaddrinfo method of Addrinfo it raises an error of SocketError
+      class SocketException < Neo4jException
+      end
+    end
+  end
+end
```
klobuczek: let's don't introduce new Neo4j exception but just catch SocketError below

ruby/neo4j/driver/internal/cluster/rediscovery_impl.rb
```diff
@@ -49,6 +49,7 @@ def resolve
             resolve_all_by_domain_name(server_address).unicast_stream
             # rescue java.net.UnknownHostException => e
           rescue => e
+            e = e.is_a?(SocketError) ? Exceptions::SocketException.new : e
             exception&.add_suppressed(e)
```
klobuczek:
```suggestion
          rescue SocketError => e
```
let's just store the 1st exception. There is not much info in those exceptions anyway.

spec/neo4j/driver/dev_manual_examples_spec.rb
```diff
@@ -34,8 +34,8 @@
 
     context 'Example 2.3. Custom Address Resolver' do
```
klobuczek: let's add something like
```
      let(:uri) { 'neo4j://nonexisting' }
```
so the spec is effective regardless of the global scheme.

klobuczek (changes requested):

lib/neo4j/driver/exceptions/socket_exception.rb
Kinju011: okay

klobuczek:
@Kinju011 the spec in this PR was not testing properly, that's why the spec was passing. The spec should be passing in the current form after my modification. Unfortunately this is not happening even for jruby.

klobuczek:
had to move the spec cluster spec as the resolver is effective only in cluster

klobuczek (approved):

PR #199, opened 2022-09-15
4.4 use connection pool gem part 2

mrhardikjoshi:


mrhardikjoshi:
closing this as #197 is better solution compared to this.

PR #200, opened 2022-09-19
fix for invalid bookmark test

Kinju011:


ruby/neo4j/driver/internal/cluster/routing_procedure_cluster_composition_provider.rb
```diff
@@ -24,7 +24,7 @@ def get_cluster_composition(connection, database_name, bookmark, impersonated_us
 
         def process_routing_response(response)
           unless response.success?
-            raise response.error.class, "Failed to run '#{invoked_procedure_string(response)}' on server. Please make sure that there is a Neo4j server or cluster up running."
+            raise response.error, "Failed to run '#{invoked_procedure_string(response)}' on server. Please make sure that there is a Neo4j server or cluster up running."
```
klobuczek, line 27: given this is shorter, but I think it does not make any difference

klobuczek: ok, it could make a difference if code is set

Kinju011: looking into the async_result method

PR #206, opened 2022-10-31
Failing spec for interrupts and exceptions

mrhardikjoshi:


ruby/neo4j/driver/internal/async/network_connection.rb
```diff
@@ -16,7 +16,7 @@ def initialize(channel, channel_pool, logger, &on_pool_shutdown)
           @server_version = channel.attributes[:server_version]
           @protocol = Messaging::BoltProtocol.for_channel(channel)
           @channel_pool = channel_pool
-          @on_pool_shutdow = on_pool_shutdown
+          @on_pool_shutdown = on_pool_shutdown
```
mrhardikjoshi, line 19: Fixed typo here and below

mrhardikjoshi, line 19: Fixed typo here and below

mrhardikjoshi (commented):
TO DO: This spec doesn't pass in jruby because the `DriverInternalDataAccessor` was written considering CRuby implementation of this gem. it needs to be updated to work with jruby as well

PR #218, opened 2022-11-28
v5 protocol implemented

Kinju011:


georgkreimer:
Are there any plans on releasing v5 support?

PR #228, opened 2023-03-13
Fix logging calls to old method.

boone:
It looks like the `trace_or_debug` method was removed in #129, and replaced with a `debug` method. I found a few references to the old method and updated them in this PR.


klobuczek (approved):

klobuczek:
@boone thank you for the PR.

PR #233, opened 2024-10-10
Fix issues with nested Sets

petergebala:
Hi,

Code there is quite complex I am not 100% convinced that my solution is correct.

```ruby
        def fresh_cluster_composition_fetched(composition_lookup_result)
          @log.debug("Fetched cluster composition for database '#{@database_name.description}'. #{composition_lookup_result.cluster_composition}")
          @routing_table.update(composition_lookup_result.cluster_composition)
          @routing_table_registry.remove_aged
          addresses_to_retain = @routing_table_registry.all_servers.map(&:unicast_stream).reduce(&:+)

          composition_lookup_result.resolved_initial_routers&.then do |addresses|
            addresses_to_retain << addresses
          end

          @connection_pool.retain_all(addresses_to_retain)

          @log.debug("Updated routing table for database '#{@database_name.description}'. #{routing_table}")
          @routing_table
        rescue => error
          cluster_composition_lookup_failed(error)
        end
```

I think problem is with `<<` method here. Looks like both lines:
```ruby
@routing_table_registry.all_servers.map(&:unicast_stream).reduce(&:+)
```
```ruby
composition_lookup_result.resolved_initial_routers
```
Return a `Set`:
```ruby
irb(main):152> composition_lookup_result.resolved_initial_routers
=> #<Set: {#<Neo4j::Driver::Internal::BoltServerAddress:0x00007f39d5a8eb78 @connection_host="XXXXXX", @host="XXXXXX", @port=7687>}>
irb(main):153> ActiveGraph::Base.driver.session_factory.connection_provider.routing_table_registry.all_servers.map(&:unicast_stream).reduce(&:+)
=>
#<Set:
 {#<Neo4j::Driver::Internal::BoltServerAddress:0x00007f39d00503a0 @connection_host="XXXXXX", @host="XXXXXX", @port=7687>,
  #<Neo4j::Driver::Internal::BoltServerAddress:0x00007f39d0050080 @connection_host="XXXXXX", @host="XXXXXX", @port=7687>,
  #<Neo4j::Driver::Internal::BoltServerAddress:0x00007f39d0050710 @connection_host="XXXXXX", @host="XXXXXX", @port=7687>}>
```

By using `<<` you create nested structure:
```ruby
irb(main):001> a = Set.new
=> #<Set: {}>
irb(main):002> b = Set.new
=> #<Set: {}>
irb(main):003> a << 1
=> #<Set: {1}>
irb(main):004> b << 2
=> #<Set: {2}>
irb(main):005> a << b
[...]

PR #235, opened 2025-03-18
Upgrade async gem, fix specs, dockerize project, make ruby 3.4 compatible

petergebala:
Hi @klobuczek 

Referring back to your comment: https://github.com/neo4jrb/neo4j-ruby-driver/pull/233#issuecomment-2407921432 

I saw you were locking `Async` gem version and trying to have green build.
I was playing a bit with neo4j-ruby-driver locally. Looks like in Async 2.13 version they changed `Queue` implementation.
Now `Queue` does not inherit from `Notification` but it is separate object that you can pass to constructor:
https://github.com/socketry/async/compare/v2.12.0...v2.13.0#diff-df8aafb695fd1c0346bb5904ac8a467d022fa17e9f018d3e337f4bc7cce0abe8R16-R21

Besides changes to the code I added Dockerfile and docker-compose for easier setup. 
I guess future contributors wouldn't like to install specific version of java or neo4j to test certain scenarios on their machines. 

Also `csv` is no longer a part of stdlib in ruby 3.4. It is a separate gem. 

Please let me know if code looks ok. I am happy to change it according to your guidance.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM ruby:3.4
+
+ARG JAVA_VERSION
+ENV JAVA_VERSION=$JAVA_VERSION
+
+# Adding Java package
+RUN apt install -y wget apt-transport-https gpg
+RUN wget -qO - https://packages.adoptium.net/artifactory/api/gpg/key/public | gpg --dearmor | tee /etc/apt/trusted.gpg.d/adoptium.gpg > /dev/null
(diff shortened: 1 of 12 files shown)


klobuczek:
@petergebala thank you for the PR. However, I'm not that keen on maintaining the 4.4 version for compatibility with newer ruby and gem versions. Could you redo the PR on 5.0 branch. We use that branch and beta driver version in production on a massive project. It should support everything 4.4 does and more. The reason it has not been released is that it does not support all the features of official ver 5 drivers and does not pass the full testkit suite. The corresponding activegraph version is 12.
We have as well given up on the cluster tests and favor testkit which is an order of magnitude more comprehensive as tests are written by neo4j staff themselves.

PR #236, opened 2025-03-19
Upgrade async gem fix specs 5.0

petergebala:
Hi @klobuczek 

As you mentioned in https://github.com/neo4jrb/neo4j-ruby-driver/pull/235
Here you have code changes against the `5.0` branch.

About cluster tests and Testkit. 
If I understand correctly, the ultimate goal is to eliminate cluster integration tests, that are setup with `neoctrl-*`.
What about other Ruby specs in the main project? Do you want to keep them? Is Testkit gonna be the only source of truth?

I haven't run Testkit yet in the docker container. I am not sure what is your point of view of having dockerized version.

Waiting for your comments :)


Rakefile
```diff
@@ -47,10 +49,13 @@ end.spec 'neo4j-ruby-driver' do
     spec_extras[:requirements] = ->(requirements) { requirements << 'jar org.neo4j.driver, neo4j-java-driver-all, 5.27.0' }
     spec_extras[:platform] = 'java'
   else
-    dependency 'async', '< 2.13'
+    dependency 'async', '>= 0'
```
klobuczek: do your changes work with async < 2.13?

klobuczek:
@petergebala sorry for the delayed response. 

> Hi @klobuczek
> 
> As you mentioned in #235 Here you have code changes against the `5.0` branch.
> 
> About cluster tests and Testkit. If I understand correctly, the ultimate goal is to eliminate cluster integration tests, that are setup with `neoctrl-*`. 

It is not so much a goal to eliminate the cluster tests as it is a pain to maintain them. They are quite brittle and do not work on driver 5.0 at all.

> What about other Ruby specs in the main project? Do you want to keep them? 

Yes, keep all tests that run against a single neo4j instance or no instance at all.

> Is Testkit gonna be the only source of truth?

Not the only source but having them all pass would be fantastic.

> 
> I haven't run Testkit yet in the docker container. I am not sure what is your point of view of having dockerized version.

testkit on github actions runs in docker container. Some of the testkit tests run against a stubbed server, which can easily simulate different scenarios that would be hard to do on a real server.
On local I run neo4j in docker, but then for debugging purposes, I run the testkit-backend and the python script directly in shell. I haven't gotten yet to a point where everything passes, so there is always debugging involved.
> 
> Waiting for your comments :)

Rakefile
petergebala: Probably not. I will adjust PR.

petergebala:
I pushed changes. I was also able to run testkit locally and replicate failing tests. I will work on them in free time.

klobuczek:
@petergebala could you refresh your branch with the current 5.0. We will try to merge your PR as the failing testkit is not related.

spec/integration/load_csv_spec.rb
```diff
@@ -5,7 +5,7 @@
 
 RSpec.describe 'LoadCsv', csv: true do
   let(:iris_class_names) { %w[Iris-setosa Iris-versicolor Iris-virginica] }
-  let(:file) { Tempfile.new('', 'tmp') }
+  let(:file) { Tempfile.new('', '/tmp') }
```
klobuczek, line 8: @petergebala, what prompted you to make this change? I was hoping this would fix the spec on GH Actions, but instead, it made it fail in my local environment, where it was passing previously. 
This spec is excluded from RSpec runs on GH Actions.

Issue #247, opened 2025-05-12
Session#execute_read and Session#execute_wite on jruby


klobuczek:
Implement the methods Session#execute_read and Session#execute_wite within jruby driver. Provide adequate unit tests.

PR #251, opened 2025-05-14
Session execute_read/write implementation

JuliaKovtun:
[#247](https://github.com/neo4jrb/neo4j-ruby-driver/issues/247)

### DIFF
--- a/jruby/neo4j/driver/ext/internal_session.rb
+++ b/jruby/neo4j/driver/ext/internal_session.rb
@@ -5,7 +5,6 @@ module Neo4j
     module Ext
       module InternalSession
         extend AutoClosable
-        extend Gem::Deprecate
         include ConfigConverter
         include ExceptionCheckable
         include RunOverride
@@ -19,6 +18,8 @@ module Neo4j
         %i[read write].each do |mode|
           ["#{mode}_transaction", "execute_#{mode}"].each do |method_name|
             define_method(method_name) do |**config, &block|
+              log_deprecation_warning(method_name, mode) if method_name.include? 'transaction'
+
               check do
                 super(
                   ->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) },
@@ -28,10 +29,6 @@ module Neo4j
             end
           end
         end
-
-        # TODO: Specify the date when the method will be removed
-        deprecate :read_transaction, "InternalSession#execute_read", 2026, 01
-        deprecate :write_transaction, "InternalSession#execute_write", 2026, 01
 
         # end work around
 
@@ -44,6 +41,13 @@ module Neo4j
 
         def begin_transaction(**config)
           check { super(to_java_config(Neo4j::Driver::TransactionConfig, **config)) }
+        end
+
+        private
+
+        def log_deprecation_warning(method_name, mode)
+          @deprecator ||= ActiveSupport::Deprecation.new('6.0', 'neo4j-ruby-driver')
+          @deprecator.deprecation_warning(method_name, "execute_#{mode}".to_sym)
         end
       end
     end
--- a/jruby/neo4j/driver/ext/internal_session.rb
+++ b/jruby/neo4j/driver/ext/internal_session.rb
@@ -5,6 +5,7 @@ module Neo4j
     module Ext
       module InternalSession
         extend AutoClosable
+        extend Gem::Deprecate
         include ConfigConverter
         include ExceptionCheckable
         include RunOverride
@@ -20,13 +21,17 @@ module Neo4j
             define_method(method_name) do |**config, &block|
               check do
                 super(
-                  ->(tx) { binding.pry; Struct::Wrapper.new(reverse_check { block.call(tx) }) },
+                  ->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) },
                   to_java_config(Neo4j::Driver::TransactionConfig, **config)
                 ).object
               end
             end
           end
         end
+
+        # TODO: Specify the date when the method will be removed
+        deprecate :read_transaction, "InternalSession#execute_read", 2026, 01
+        deprecate :write_transaction, "InternalSession#execute_write", 2026, 01
 
         # end work around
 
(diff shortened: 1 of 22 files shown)


jruby/neo4j/driver/ext/internal_session.rb
```diff
@@ -14,10 +14,20 @@ module InternalSession
...
+              if method_name.include? 'transaction'
+                Neo4j::Driver::Internal::Deprecator.log_warning(method_name, "execute_#{mode}".to_sym, '6.0')
+              end
+
+              check do
+                super(
+                  ->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) },
+                  to_java_config(Neo4j::Driver::TransactionConfig, **config)
+                ).object
+              end
```
klobuczek: I would extract this to a separate private method. So the above logic can be simplified.

jruby/neo4j/driver/ext/internal_session.rb
```diff
@@ -14,10 +14,20 @@ module InternalSession
...
 
-        %i[read write].each do |prefix|
-          define_method("#{prefix}_transaction") do |**config, &block|
-            check do
-              super(->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) }, to_java_config(Neo4j::Driver::TransactionConfig, **config)).object
+        # implementation of read_transaction, write_transaction, execute_read, execute_write
+        %i[read write].each do |mode|
+          ["#{mode}_transaction", "execute_#{mode}"].each do |method_name|
+            define_method(method_name) do |**config, &block|
+              if method_name.include? 'transaction'
```
klobuczek: This is kind of backwards and sort of duplication.
I think in this case it would be better to have 2 separate method definitions. One with deprecation and another without.

jruby/neo4j/driver/ext/internal_session.rb
```diff
@@ -14,10 +14,20 @@ module InternalSession
...
+                Neo4j::Driver::Internal::Deprecator.log_warning(method_name, "execute_#{mode}".to_sym, '6.0')
+              end
+
+              check do
+                super(
+                  ->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) },
+                  to_java_config(Neo4j::Driver::TransactionConfig, **config)
+                ).object
```
klobuczek: let's increase our max line length to 160. That's what we have on all our projects.

jruby/neo4j/driver/ext/internal_session.rb
```diff
@@ -14,10 +14,20 @@ module InternalSession
...
+        %i[read write].each do |mode|
+          ["#{mode}_transaction", "execute_#{mode}"].each do |method_name|
+            define_method(method_name) do |**config, &block|
+              if method_name.include? 'transaction'
+                Neo4j::Driver::Internal::Deprecator.log_warning(method_name, "execute_#{mode}".to_sym, '6.0')
+              end
+
+              check do
+                super(
+                  ->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) },
```
klobuczek: Please check if Struct::Wrapper is still required.

klobuczek (changes requested):
Our spec `write transaction function retries deadlocks` corresponds to java test `writeTransactionFunctionShouldRetryDeadlocks`. That test is marked as deprecated. I don't see any tests testing the same functionality with `executeWrite` in java so we should not either. Let's exclude this test.
Furthermore, I think it would make sense to disable the deprecation message in test environment for readability.

jruby/neo4j/driver/ext/internal_session.rb
JuliaKovtun: okay, will check

jruby/neo4j/driver/ext/internal_session.rb
```diff
@@ -14,11 +14,15 @@ module InternalSession
...
-            end
+        # implementation of read_transaction, write_transaction, execute_read, execute_write
+        %i[read write].each do |mode|
+          define_method("execute_#{mode}") do |**config, &block|
+            execute_transaction(__method__, **config, &block)
+          end
+
+          define_method("#{mode}_transaction") do |**config, &block|
+            Neo4j::Driver::Internal::Deprecator.log_warning("#{mode}_transaction", "execute_#{mode}".to_sym, '6.0')
+            execute_transaction(__method__, **config, &block)
```
klobuczek: ouch! That got complex. I didn't realize that when making the recommendation.
Seeing that I would prefer to duplicate the method implementation or go back to the conditional deprecation you had until we remove the method.

jruby/neo4j/driver/ext/internal_session.rb
```diff
@@ -14,11 +14,15 @@ module InternalSession
...
-              super(->(tx) { Struct::Wrapper.new(reverse_check { block.call(tx) }) }, to_java_config(Neo4j::Driver::TransactionConfig, **config)).object
-            end
+        # implementation of read_transaction, write_transaction, execute_read, execute_write
+        %i[read write].each do |mode|
+          define_method("execute_#{mode}") do |**config, &block|
+            execute_transaction(__method__, **config, &block)
+          end
+
+          define_method("#{mode}_transaction") do |**config, &block|
+            Neo4j::Driver::Internal::Deprecator.log_warning("#{mode}_transaction", "execute_#{mode}".to_sym, '6.0')
```
klobuczek: You can use relative namespace
```suggestion
            Internal::Deprecator.log_warning("#{mode}_transaction", "execute_#{mode}".to_sym, '6.0')
```

klobuczek (changes requested):

jruby/neo4j/driver/ext/internal_session.rb
JuliaKovtun: I tried `Internal::Deprecator.log_warning` but it raised a NameError, so left here `Driver::Internal::Deprecator.log_warning`

klobuczek (approved):

PR #253, opened 2025-05-16
Add examples for features not yet implemented

### LINKED ISSUE #239: Documentation
It would be great to have documentation similar to https://neo4j.com/docs/java-manual/current/ from `Quickstart` to `Data types and mapping to Cypher types`

nsauk:


docs/dev_manual_todo_examples.rb
```diff
@@ -0,0 +1,179 @@
...
+# Filter notifications
+# Allow only WARNING notifications, but not of HINT or GENERIC classifications
+driver = Neo4j::Driver::GraphDatabase.driver(
+  'bolt://localhost:7687',
+  Neo4j::Driver::AuthTokens.basic('neo4j', 'password'),
+  notification_config: {
+    minimum_severity: :warning, # use :off to disable entirely
+    disabled_categories: [:hint, :generic]
+  }
+)
```
klobuczek: I think you should have at most one or two examples specifying the AuthToken and QueryConfig. Most examples should skip those parameters to show case the most common scenario (I believe 99%) of query and parameters only.

klobuczek (changes requested):

docs/dev_manual_todo_examples.rb
nsauk: Done

klobuczek (approved):

PR #254, opened 2025-05-20
Fix notification config

nsauk:


jruby/neo4j/driver/ext/config_converter.rb
```diff
@@ -70,7 +70,7 @@ def notification_config(minimum_severity: nil, disabled_categories: nil)
           org.neo4j.driver.internal.InternalNotificationConfig.new(
             value_of(org.neo4j.driver.internal.InternalNotificationSeverity, minimum_severity),
             disabled_categories
-              &.map { |value| value_of(org.neo4j.driver.internal.InternalNotificationCategory, value) }
+              &.map { |value| org.neo4j.driver.NotificationCategory.const_get(value.to_s.upcase) }
```
klobuczek: We have been dealing with this issue the following way: https://github.com/neo4jrb/neo4j-ruby-driver/blob/5.0/ruby/neo4j/driver/access_mode.rb
to have some level of stronger typing compatible with java.
I could be swayed over but could end up with complex converting code when the configuration is deeply nested resulting in runtime errors as users would be using nested symbol hashes instead of higher level types.

klobuczek: Using NotificationClassification serves as well as a type of documentation.

nsauk:
Converted to draft because the spec is incorrect

jruby/neo4j/driver/ext/config_converter.rb
```diff
@@ -67,15 +67,22 @@ def trust_strategy(**config)
         end
 
         def notification_config(minimum_severity: nil, disabled_categories: nil)
-          org.neo4j.driver.internal.InternalNotificationConfig.new(
-            value_of(org.neo4j.driver.internal.InternalNotificationSeverity, minimum_severity),
-            disabled_categories
-              &.map { |value| value_of(org.neo4j.driver.internal.InternalNotificationCategory, value) }
```
klobuczek, line 73: @nsauk could you check. It appears to me that the only change needed was this line:
```
&.map { |value| value_of(org.neo4j.driver.NotificationClassification, value) }
```

klobuczek (changes requested):

jruby/neo4j/driver/ext/config_converter.rb
klobuczek: Maybe, additionally move the `.or_else(nil)` from `value_of` method to the line 71 as the `value_of` of java enum, which NotificationClassification now is, does not return `Optional`.

nsauk: I rewrote the rest of this method using the Java driver's public API so we rely less on internals.

InternalNotificationSeverity works with `value_of`, but NotificationSeverity fails because it's an interface. As far as I can understand, we can't use the same approach for both (unless it's `const_get`).

Issue #238, opened 2025-04-24
Improve README


klobuczek:
Improve the README.md in the style of https://github.com/neo4j/neo4j-java-driver/blob/5.0/README.md

PR #255, opened 2025-05-20
Fix metadata for Hoe

nsauk:
In #238 I removed a line that was used by Hoe to set the gem's metadata, therefore it fails on rake tasks.

I was unable to quickly find a way to put this into the Hoe config (`self.homepage = ` didn't work), so this PR just adds the line back. Double whitespace before `::` as in the previous version of the README aren't required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Neo4j Ruby Driver
 
+home :: https://github.com/neo4jrb/neo4j-ruby-driver
+
 This repository contains 2 implementation of a Neo4j driver for Ruby:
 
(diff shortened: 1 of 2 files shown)


nsauk:
About the latest commit: the 'home' key in the `urls` hash has to be a string because of https://github.com/seattlerb/hoe/blob/b80a4958591d7b7cdfbf6c9e743f76e6558f6dac/lib/hoe.rb#L549, otherwise it fills s.homepage but not s.metadata["homepage_uri"] in .gemspec

klobuczek (approved):

PR #256, opened 2025-05-23
Wrapper for execute_read/write transaction

JuliaKovtun:
Wrapper for execute_read/write (mri implementation)  #251


ruby/neo4j/driver/internal/delegating_transaction.rb
```diff
@@ -0,0 +1,15 @@
...
+module Neo4j::Driver
+  module Internal
+    class DelegatingTransaction
+      def initialize(tx)
+        @tx = tx
+      end
+
+      def run(query, **parameters)
+        @tx.run(query, **parameters)
+      end
```
klobuczek: Use `delegate :run, to: :@tx` at the top of the class definition

ruby/neo4j/driver/internal/internal_session.rb
```diff
@@ -48,11 +48,11 @@ def write_transaction(**config, &block)
       end
 
       def execute_read(**config, &block)
-        transaction(AccessMode::READ, **config, &block)
+        transaction(AccessMode::READ, **config, &(->(tx) { block.call(DelegatingTransaction.new(tx)) }))
```
klobuczek: We have a nicer more idiomatic syntax for that,
```suggestion
        transaction(AccessMode::READ, **config) { block.call(DelegatingTransaction.new(it)) }
```
Alternatively composition,  bonus points for investigation. Personally never used it before:
```suggestion
        transaction(AccessMode::READ, **config, &(DelegatingTransaction.method(:new) >> block))
```
Looking at it I don't think it beats the former basic ruby syntax.

klobuczek (changes requested):

ruby/neo4j/driver/internal/internal_session.rb
JuliaKovtun: I went with the first option - it's simple and readable. I also tested the second (with >>), and it works well too, thanks for the suggestions

testkit-backend/lib/testkit/backend/messages/requests/session_read_transaction.rb
```diff
@@ -1,6 +1,9 @@
...
   module Requests
     class SessionReadTransaction < SessionTransaction
+      def process
+        super(:execute_read)
+      end
     end
   end
-end
\ No newline at end of file
+end
```
klobuczek, line 9: I think we want the final new line character. It has been suggested by rubocop and we have been following it sine then.

ruby/neo4j/driver/internal/internal_session.rb
```diff
@@ -48,11 +48,11 @@ def write_transaction(**config, &block)
       end
 
       def execute_read(**config, &block)
-        transaction(AccessMode::READ, **config, &block)
+        transaction(AccessMode::READ, **config) { |tx| block.call(DelegatingTransaction.new(tx)) }
       end
 
       def execute_write(**config, &block)
-        transaction(AccessMode::WRITE, **config, &block)
+        transaction(AccessMode::WRITE, **config) { |tx| block.call(DelegatingTransaction.new(tx)) }
```
klobuczek: Now that it has been simplified, it becomes more obvious that the duplication needs to be fixed.
Introduce a parameter in `transaction` method or maybe better a private `delegating_transaction` method which delegates to `transaction` with modified block.

klobuczek (changes requested):

testkit-backend/lib/testkit/backend/messages/requests/session_read_transaction.rb
JuliaKovtun: Yes, makes sense - I added the final new line, which removed the warning. GitHub just doesn't show it in the diff.

klobuczek (approved):

PR #258, opened 2025-06-02
Add execute_query

nsauk:


jruby/neo4j/driver/ext/internal_driver.rb
```diff
@@ -11,6 +11,19 @@ module InternalDriver
...
+          java_method(:executableQuery, [java.lang.String])
+            .call(query)
+            .java_method(:withParameters, [java.util.Map])
+            .call(parameters.transform_keys(&:to_s))
+            .java_method(:withAuthToken, [org.neo4j.driver.AuthToken])
+            .call(auth_token)
+            .java_method(:withConfig, [org.neo4j.driver.QueryConfig])
+            .call(to_java_config(Neo4j::Driver::QueryConfig, **config))
+            .java_method(:execute, [])
+            .call
```
klobuczek: you will need to wrap the entire method with 
```
check do
...
end
```
which converts potential exceptions from java to ruby exceptions.

jruby/neo4j/driver/ext/internal_driver.rb
```diff
@@ -11,6 +11,19 @@ module InternalDriver
 
         auto_closable :session
 
+        def execute_query(query, auth_token = nil, config = {}, **parameters)
+          java_method(:executableQuery, [java.lang.String])
+            .call(query)
+            .java_method(:withParameters, [java.util.Map])
+            .call(parameters.transform_keys(&:to_s))
```
klobuczek: you need to get hold of the actual java method object only in relatively seldom cases. Mostly the methods can be determined dynamically.
I'm quite sure this will work:
```suggestion
            .with_parameters(to_neo(parameters))
```
please note the conversion `to_neo`. Keys and values have to be converted.

jruby/neo4j/driver/ext/internal_driver.rb
```diff
@@ -11,6 +11,19 @@ module InternalDriver
...
+            .java_method(:withAuthToken, [org.neo4j.driver.AuthToken])
```
klobuczek: always check if you really need the explicit method lookup. The dynamic dispatch is so much more readable.

klobuczek (changes requested):

jruby/neo4j/driver/ext/internal_driver.rb
```diff
@@ -11,6 +11,16 @@ module InternalDriver
 
         auto_closable :session
 
+        def execute_query(query, auth_token = nil, config = {}, **parameters)
+          check do
+            driver.executable_query(query)
```
klobuczek: I was stumbling on the `driver` until I realized that it might refer to partial package name. We don't such references. The method `driver` in this context is confusing. Just remove it.

jruby/neo4j/driver/ext/internal_driver.rb
```diff
@@ -11,6 +11,16 @@ module InternalDriver
...
+              .with_parameters(to_neo(parameters))
+              .with_auth_token(auth_token)
+              .with_config(to_java_config(Neo4j::Driver::QueryConfig, **config))
```
klobuczek: Let's keep the order of calls the same as the parameter order.

klobuczek (changes requested):

testkit-backend/lib/testkit/backend/messages/responses/eager_result.rb
```diff
@@ -0,0 +1,34 @@
...
+
+          include SummaryHelper
+
+          def to_testkit(*args)
+            if args.empty?
+              response_to_testkit
+            else
+              value_to_testkit(*args)
+            end
+          end
```
nsauk: At the moment I managed to make it work with this trick only.

Issue #245, opened 2025-05-05
Add missing teskit-backend requests


klobuczek:
Add all missing Testkit::Backend::Messages::Requests according to testkit/nutkit/requests.
Try implementing the requests, they don't have to run and let's sync.

PR #260, opened 2025-06-04
ResultSingleOptional testkit request

JuliaKovtun:
ResultSingleOptional testkit request implementation (#245)

### DIFF
--- a/testkit-backend/lib/testkit/backend/messages/requests/result_single_optional.rb
+++ b/testkit-backend/lib/testkit/backend/messages/requests/result_single_optional.rb
@@ -2,12 +2,8 @@ module Testkit::Backend::Messages
   module Requests
     class ResultSingleOptional < Request
       def process
-        result = fetch(result_id)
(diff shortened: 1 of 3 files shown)


testkit-backend/lib/testkit/backend/messages/requests/result_single_optional.rb
```diff
@@ -0,0 +1,15 @@
+module Testkit::Backend::Messages
+  module Requests
+    class ResultSingleOptional < Request
+      def process
+        result = fetch(result_id)
+        record = begin
+          { values: result.single.values.map(&method(:to_testkit)) }
+        rescue Neo4j::Driver::Exceptions::NoSuchRecordException
```
klobuczek: Using exception for control flow is not  a good practice. The same can be achieved without raising and rescuing an exception by reading 2 records.

klobuczek (changes requested):

PR #267, opened 2026-01-16
Fix TimedStack initialization

nsauk:
In connection_pool 3.0.0 they now use keyword arguments rather than positional arguments everywhere.
https://github.com/mperham/connection_pool/blob/main/Changes.md#300

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -2,9 +2,9 @@ name: Specs
 
 on:
   push:
-    branches: [ 5.0 ]
+    branches: [ 5.0, 6.x ]
   pull_request:
-    branches: [ 5.0 ]
+    branches: [ 5.0, 6.x ]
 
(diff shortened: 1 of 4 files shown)


nsauk:
Triggered specs manually: https://github.com/neo4jrb/neo4j-ruby-driver/actions/runs/21072304694

nsauk:
And once again: https://github.com/neo4jrb/neo4j-ruby-driver/actions/runs/21072771653

klobuczek:
I assume this started occurring with a gem upgrade. After you adjusted to the newer version you should put lower bound on that gem version in Rakefile so people won't be able to use the driver with older version of connection pool.

nsauk:
https://github.com/neo4jrb/neo4j-ruby-driver/actions/runs/21084907909

nsauk:
@klobuczek I think it's enough, please look again

klobuczek (approved):

PR #275, opened 2026-02-26
Fix delegating transaction

nsauk:
This will allow use of 6.x in dependent projects.


ruby/neo4j/driver.rb
```diff
@@ -15,6 +15,7 @@ module Neo4j
   module Driver
     Loader.load
 
+    DelegatingTransaction = Neo4j::Driver::Internal::DelegatingTransaction
```
klobuczek: Why is this necessary?

klobuczek:
Could you create a failing spec to understand the issue before the change?

klobuczek (approved):

klobuczek:
Deployed [6.0.0.alpha.2](https://rubygems.org/gems/neo4j-ruby-driver/versions/6.0.0.alpha.2)

PR #288, opened 2026-05-06
Correct cross-flavor override doc — per-gem force_ruby_platform

klobuczek:
## Summary
JRUBY.md and the loader's comment claimed there was no clean per-gem override for the cross-flavor case (MRI flavor on JRuby). That was wrong — Bundler's Gemfile DSL accepts \`force_ruby_platform: true\` as a per-gem option that scopes the platform flip to that one dep, leaving everything else to resolve natively:

\`\`\`ruby
gem 'neo4j-ruby-driver', force_ruby_platform: true
\`\`\`

This works for any source (rubygems, path, git), so it's the canonical way for a JRuby user to opt into the MRI flavor.

## Diff
- JRUBY.md "Selecting a flavor" section now leads with the per-gem option, with \`glob:\` / \`gemspec name:\` as equivalent alternatives for path sources.
- \`lib/shared/neo4j/driver.rb\` impl-detection comment updated to reference the per-gem option, not the (overstated) "pins gemspec discovery" workaround.

## Out of scope
The CI does not yet exercise the MRI-on-JRuby flavor; we deferred that in #283. With this Bundler option known to work, adding such a row is realistic later.

### DIFF
--- a/JRUBY.md
+++ b/JRUBY.md
@@ -147,20 +147,34 @@ chosen impl and branches on `STAGED_BUILD`:
   a `path:` source.
 - Staged: files = flat `lib/**/*`, `require_paths = ['lib']`.
 
-### Selecting a flavor from source
+### Selecting a flavor
 
-Consumer Gemfile (path or git source):
+Default consumer Gemfile (any source):
 
 ```ruby
-gem 'neo4j-ruby-driver', path: '../neo4j-ruby-driver'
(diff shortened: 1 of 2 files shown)


klobuczek:
Superseded by #289, which incorporates the same doc fix plus the actual implementation (Gemfile env-var bridge + CI matrix row).

PR #305, opened 2026-05-18
Routing slice 8: strip empty values from BEGIN extras

klobuczek:
## Why

The biggest cluster of RoutingV4x4 errors left after slice 7 (~25 distinct tests × 4 parameter sets) was a single script-mismatch on `BEGIN`:

```
Expected: BEGIN {"db": "adb"}
Received: BEGIN {"db": "adb", "tx_metadata": {}}
```

`transaction.rb` built the BEGIN extras with `begin_extra[:tx_metadata] = options[:metadata] if options[:metadata]` — which is truthy for an empty hash, so `tx_metadata: {}` always ended up on the wire. The Rust boltstub scripts match the BEGIN map strictly.

## What's in

Replace the per-key conditional inserts with the same `reject!` pattern session.rb already uses for RUN extras: build the map with raw values, then drop entries whose value is nil or an empty collection. Matches what every other driver puts on the wire when the session/tx didn't set metadata or bookmarks.

## Test plan

- [x] `bundle exec rspec spec/shared/neo4j/driver` (207 / 0)
- [ ] testkit-stub baseline gate green
- [ ] Expect ~25 RoutingV4x4 tests (× parameter variants) to flip from ERROR to OK


lib/mri/neo4j/driver/transaction.rb
```diff
@@ -18,12 +18,19 @@ def initialize(connection, session, bookmarks = [], options = {}, on_release: ni
...
+        # expect (e.g. `BEGIN {"db": "adb"}` not
+        # `BEGIN {"db": "adb", "tx_metadata": {}}`).
+        begin_extra = {
+          bookmarks: bookmarks,
+          db: options[:database],
+          mode: options[:access_mode],
+          tx_timeout: options[:timeout],
+          tx_metadata: options[:metadata]
+        }
+        begin_extra.reject! { |_, v| v.nil? || (v.respond_to?(:empty?) && v.empty?) }
```
klobuczek: We have the `blank?` implementation in multiple places. Deduplicate.

PR #332, opened 2026-05-27
spec(parameters): port missing tests from neo4j-java-driver ParametersIT

klobuczek:
Fourth IT-port PR (after #323 SessionIT, #330 TransactionIT, #331 ErrorIT).

### Ported (7, all passing)

| Java method | rspec |
|---|---|
| `shouldSendAndReceiveLongString` | `sends and receives a long string round-trip` |
| `shouldSendAndReceiveLongListOfLongs` | `sends and receives a long list of integers round-trip` |
| `shouldSendAndReceiveLongArrayOfBytes` | `sends and receives a long byte array round-trip` |
| `shouldNotBePossibleToUseNodeAsParameterInMapValue` | `rejects a Node as a parameter` |
| `shouldNotBePossibleToUseRelationshipAsParameterViaMapValue` | `rejects a Relationship as a parameter` |
| `shouldNotBePossibleToUsePathAsParameterViaMapValue` | `rejects a Path as a parameter` |
| `settingInvalidParameterTypeShouldThrowHelpfulError` | `rejects an arbitrary object as a parameter with a helpful error` |

### Not ported (with reasons)
- **`shouldBeAbleToSetAndReturn{Byte,Short,Long}Property`** — Java primitive width distinctions don't apply to Ruby (`Integer` covers all integral widths); the existing `Integer` / `Integer Array` examples already cover round-trip.
- **`settingInvalidParameterTypeDirectlyShouldThrowHelpfulError`** — exercises a Java-specific `session.run(query, NodeValue)` signature where the second positional arg is itself a `Value` (not a `Map`); Ruby's `session.run` only accepts a Hash / kwargs there.

### Note on assertion shape
The Ruby driver routes all unsupported-type-as-parameter rejections through the *same* path (`"Unable to convert X to Neo4j Value."`) — it has no `MapValue` wrapper that would produce Java's distinct `"Unknown type: NODE"`. The Node/Rel/Path examples therefore assert on the unified Ruby message plus a type-name match, which is the stable cross-impl shape.

### Verification
Full `parameters_spec.rb`: **30 examples, 0 failures** (was 23).

### DIFF
--- a/spec/shared/integration/parameters_spec.rb
+++ b/spec/shared/integration/parameters_spec.rb
@@ -192,4 +192,44 @@ RSpec.describe 'Parameters' do
     received_value = session.run('RETURN $value', value: stream).single[0]
     expect(received_value).to eq [1, 2, 3, 4, 5, 42]
   end
+
+  # --- Ported from neo4j-java-driver ParametersIT.java ---
+
+  it 'sends and receives a long string round-trip' do
+    long_string = 'x' * 65_536
+    expect(session.run('RETURN $value', value: long_string).single[0]).to eq long_string
+  end
+
+  it 'sends and receives a long list of integers round-trip' do
+    longs = Array.new(65_536) { rand(0..2**31 - 1) }
+    expect(session.run('RETURN $value', value: longs).single[0]).to eq longs
+  end
+
+  it 'sends and receives a long byte array round-trip' do
+    bytes = (Array.new(65_536) { rand(0..255).chr }.join).b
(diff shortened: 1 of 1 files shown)


klobuczek:
Self-close: re-reading parameters_spec.rb more carefully showed the existing `context 'invalid parameter types'` (Object / Node / Relationship / Path via `shared_examples 'raises exception'` and `it_behaves_like`) and the `context 'long values'` block already cover every one of the 7 Java tests I tried to port. My initial grep only matched `^\s*it[ '"]` and missed the context-based / parameterized forms. Closing as duplicate.

PR #348, opened 2026-06-06
feat(jruby): mutual-TLS client certificates → Feature:API:SSLClientCertificate

klobuczek:
## Summary
Plan item #5. Wires mutual-TLS client certificates on JRuby and advertises `Feature:API:SSLClientCertificate` (`'ja'`).

The JRuby `DriverFactory#new_instance` wrapper already passed `nil` for the `ClientCertificateManager` arg of Java's 4-arg `newInstance` with a comment "wired in a later slice" — this is that slice.

## Changes
- **`DriverFactory#new_instance`** (JRuby) takes an optional `client_certificate_manager:` and passes it to Java's 4-arg `newInstance`, which derives the security plan from it + the `+s`/`+ssc` scheme. **MRI's factory** accepts the kwarg and ignores it (mTLS client certs are JRuby-only; testkit never sends one to MRI).
- **`Neo4j::Driver::ClientCertificateManager`** (JRuby) — a block-based wrapper that `include`s `org.neo4j.driver.ClientCertificateManager`, mirroring the `AuthTokenManager` pattern. `getClientCertificate` wraps the block result (a cert, or `nil` for "no change") in a completed future.
- **testkit `NewDriver`** builds the manager: static `clientCertificate` → `ClientCertificates.of(cert, key[, password])` wrapped in `ClientCertificateManagers.rotating` (like Java's backend); `clientCertificateProviderId` → the managed provider.
- **`NewClientCertificateProvider`** creates a real provider that round-trips to the frontend (blocking) per `getClientCertificate`, returning a fresh cert when `hasUpdate` else `nil`. **`ClientCertificateProviderCompleted`** is the inline reply (no-op, like `AuthTokenManagerGetAuthCompleted`).

## Test plan
- [x] JRuby smoke test with real testkit cert files: `ClientCertificates.of` (with/without password), `ClientCertificateManagers.rotating`, the provider wrapper's `getClientCertificate` (cert + `nil`), driver construction through the factory with a `client_certificate_manager`.
- [x] Full shared rspec green on both impls (JRuby 428/0/1, MRI 425/0/1).
- [ ] **CI testkit-tls** runs the actual mTLS handshake (`TestClientCertificate` + `TestClientCertificateRotation`) against a client-cert-requiring TLS server — the part not reproducible locally. Fold newly-passing tests into the jruby tls baseline.

## Notes
- MRI mTLS client certs are **not** implemented (the pure-Ruby Bolt TLS layer would need client-cert support); feature stays `'ja'`, not `'jar'`.

## Summary by CodeRabbit

* **New Features**
  * Enabled mutual TLS (mTLS) for JRuby with a client-certificate manager and support for certificate rotation; drivers can accept a client-certificate provider.
  * JRuby driver factory now forwards an optional client-certificate manager through driver creation (MRI accepts the argument but ignores it).

* **Tests**
  * Test tooling and baselines updated to include JRuby SSL client-certificate scenarios and certificate rotation.


lib/jruby/neo4j/driver/ext/internal/driver_factory.rb
```diff
@@ -39,11 +39,11 @@ def to_clock(clock)
             ClockAdapter.new(clock)
           end
 
-          def new_instance(uri, auth_token_manager, config = {})
+          def new_instance(uri, auth_token_manager, config = {}, client_certificate_manager: nil)
```
klobuczek: unnatural formal parameter order. Order should be the same as in super.

testkit-backend/requests/new_driver.rb
```diff
@@ -41,11 +41,35 @@ def to_object
         # Ruby resolver proc / TestkitClock — no Java refs on this
         # side, the production-side base class handles the conversion.
         resolver = method(:domain_name_resolver) if domain_name_resolver_registered
-        Internal::DriverFactoryWithDomainNameResolver.new(resolver).new_instance(uri, auth_token_manager, config)
+        Internal::DriverFactoryWithDomainNameResolver.new(resolver)
+                .new_instance(uri, auth_token_manager, config, client_certificate_manager: client_certificate_manager)
```
klobuczek: it's ugly to have config as a positional parameter, but then add a keyword parameter. config should be the last parameter to be able to specify literal keywords. client_certificate_manager should be positional or folded into config.

testkit-backend/requests/new_driver.rb
```diff
@@ -41,11 +41,35 @@ def to_object
...
+
+      def to_client_certificate(cert)
+        cert_file = java.io.File.new(cert[:certfile])
+        key_file = java.io.File.new(cert[:keyfile])
+        if cert[:password]
+          Neo4j::Driver::ClientCertificates.of(cert_file, key_file, cert[:password])
+        else
+          Neo4j::Driver::ClientCertificates.of(cert_file, key_file)
+        end
+      end
```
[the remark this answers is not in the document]
klobuczek: Not applicable: the nested fields are lowercase in the protocol (Java's ClientCertificate.Data declares String certfile/keyfile/password), and the testkit-tls run is green reading cert[:data][:certfile]. The camelCase point applies to top-level keys (which Request.from underscores), not these nested ones.

testkit-backend/requests/new_client_certificate_provider.rb
```diff
@@ -1,13 +1,41 @@
...
+        return unless completed.has_update
+
+        cert = completed.client_certificate
+        cert_file = java.io.File.new(cert[:certfile])
+        key_file = java.io.File.new(cert[:keyfile])
+        if cert[:password]
+          Neo4j::Driver::ClientCertificates.of(cert_file, key_file, cert[:password])
+        else
+          Neo4j::Driver::ClientCertificates.of(cert_file, key_file)
+        end
```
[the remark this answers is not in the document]
klobuczek: Same as the NewDriver thread: the inner cert fields are lowercase (certfile/keyfile/password per Java's ClientCertificate.Data); green in testkit-tls. No camelCase variant is sent for these nested keys.

lib/jruby/neo4j/driver/ext/internal/driver_factory.rb
klobuczek: Fixed in 71d5456 — new_instance(uri, auth_token_manager, client_certificate_manager: nil, **config): matches super's order and the explicit kwarg binds before **config so the manager can't leak into config.

testkit-backend/requests/new_driver.rb
klobuczek: Fixed in 71d5456 — config is now **kwargs (literal-keyword-able, how to_java_config(Config, **config) already consumes it) and client_certificate_manager is an explicit keyword. Callers (NewDriver, MRI GraphDatabase) splat **config.

PR #373, opened 2026-06-13
probe: forbidden-on-read-only writer tests in jruby stub baseline

klobuczek:
## Purpose — probe, not a feature

Tests the hypothesis that the routing `forbidden_on_read_only` gap is **not** driver logic. Locally on JRuby, 4 of the 5 `test_should_fail_when_writing_on_writer_that_returns_forbidden_on_read_only_database` variants pass (Java does the routing-table writer invalidation); they're just absent from the CI baseline.

Adding them to the jruby stub baseline lets CI adjudicate:
- **Green** → they pass in the Docker harness too; this is a legitimate fold (4 free wins).
- **Red** → confirms the JRuby routing *local-vs-CI* divergence (Docker timing) is the real blocker, not exception parity.

v3 is omitted — it errors locally for a separate, isolated reason (Bolt-3.0 `HELLO` sent without the `routing` context field the stub expects).

## Release Notes

* **Tests**
  * Updated test baseline entries for routing test cases to reflect expected behavior when write operations are performed on read-only databases across multiple versions.


klobuczek:
Probe complete — verdict: these 4 do NOT pass in CI. They FAIL (not error/hang) with a real assertion: the write error surfaces as SessionExpiredException with code='N/A' (gqlStatus 50N42, 'no longer accepts writes'), dropping the original Neo.ClientError.General.ForbiddenOnReadOnlyDatabase code the test asserts on. So this is an exception-code-preservation issue, not a free baseline fold and not a harness flake. Closing; tracking the real fix separately.

PR #379, opened 2026-06-14
mri(auth): complete Feature:Auth:Managed error handling

klobuczek:
Completes the hardest advertised feature with failures: **`Feature:Auth:Managed` (`AUTH_MANAGED`)**. All `tests/stub/authorization/test_auth_token_manager` tests now pass on both MRI flavors (`mri`, `mri-on-jruby`); integration and TLS suites stay green.

## 1. Notify the manager once per connection (`9928bea`)
`handle_security_exception` was firing **twice** for one error (contract: exactly once per `Security.*` failure) — the same failure was classified twice as it propagated (`Result#on_failure`, then tx rollback re-consuming the failed result).

- Dedup the notification per connection (`@security_notified`); the connection is discarded after a security failure so the flag never needs clearing.
- `AuthorizationExpired` now also goes through the manager (it's a `Security.*` error; the manager decides retryability), keeping only its RESET-ok distinction.

## 2. Resolve the per-acquire identity once, in the provider (`207ad8e`)
Build a fresh connection with the identity it should hold instead of connecting as the manager and re-authing afterwards. The session passes only its per-session override (`:auth_token`, nil = manager default); the connection provider resolves the effective token **once** and the pool's create block authenticates the new connection with it directly.

Fixes the `AuthTokenManager.get_auth_count` contract:
- direct, default identity: **1** (was 2 — `connect_factory` plus a redundant session re-auth)
- routing, default identity: **2** — ROUTE connection + worker, each its own single consult
- per-session auth (either): **0** — manager never consulted, and the connection is built with the session token so no `LOGOFF`/`LOGON` precedes `BEGIN` (matches the stub scripts)

`ensure_identity` re-auths only *reused* pooled connections, and still requires Bolt 5.1+ for per-session auth even when the popped connection already holds the token.

## 3. Fold baseline (`02133ca`)
Stub gate: **0 regressions** (MISSING=0, FAILED=0), **+8 new passes** folded into `testkit-stub-baseline-mri.txt` — the 6 `TestAuthTokenManager5x1` `test_error_on_*` cases plus `TestUserSwitchingV5x1` `test_read_with_switch_inverse{,_routing}`.

Unadvertising `Feature:Auth:Managed` would leave the total failure count unchanged (its gated tests all pass), so the feature is cleanly complete.

## Summary by CodeRabbit

* **New Features**
  * Added per-session authentication override, allowing custom auth tokens to be supplied when acquiring connections.
  * Extended identity/auth enforcement across Bolt and routing modes, including support for re-authentication when reused pooled connections are eligible.

* **Bug Fixes**
  * Improved reliability of security handling by reporting security failures to the auth-token manager at most once per connection.

* **Tests**
  * Updated baseline/stub expectations for authentication error and user-switching inverse/routing scenarios to match the new behavior.


klobuczek:
Thanks @copilot-pull-request-reviewer and CodeRabbit — all three structural findings were valid and are fixed in db0626f:

1. **`@next_auth` race** — `Bolt::Pool#pop` now hands the per-pop token to the create block via a **thread-local keyed by the pool's `object_id`** (cleared in `ensure`). The create block runs synchronously inside `@stack.pop` on the calling thread, so it's read back by the same thread that set it and concurrent pops can't clobber each other.
2. **Slot leak on `ensure_identity` failure** — both acquire paths (direct + routing worker) now `discard` the just-popped connection if identity enforcement raises (per-session auth on Bolt < 5.1, or a re-auth LOGON failure), then re-raise.
3. **Reused ROUTE connection identity** — `fetch_routing_table_from` resolves discovery's identity once and runs `ensure_identity` on the popped router connection before `ROUTE`, so a *reused* router connection authenticates as the current session's identity and honours the Bolt 5.1 gate. The fresh-case `get_token` count for routing stays at 2.

Also took the lower-confidence Copilot suggestion on `notify_security_exception`: it now **caches the classified result** so repeat classifications of the same failure return the same retryable verdict instead of downgrading to the raw exception.

MRI unit specs green; re-running the testkit suites to confirm no count regressions.

PR #381, opened 2026-06-14
mri(auth): finish Feature:Auth:Managed — session-auth notify, token rotation, authz-expired treatment

klobuczek:
Completes the `Feature:Auth:Managed` gaps that #379 left (its baseline-dedup masked ~68 parametrised auth-manager subtests). Diagnosed precisely against a **local boltstub run** (Python boltstub + MRI backend), not guessed from CI.

## Commits
1. **Session-auth notification gate** — a security failure on a per-session token must not reach the manager (`handle_security_exception_count` stays 0). Fixes 48 `test_error_on_*` `session_auth=True` FAILs.
2. **Bolt 5.0 stale-token replacement (direct)** — 5.0 has no LOGOFF/LOGON, so on token rotation a reused connection is discarded and rebuilt under the new identity.
3. **Authorization-expired pool treatment** — root cause of the renewing/handles_unknown ERRORs: when one connection hits `AuthorizationExpired`, the server drops its authorization cache for that identity *on that server*, so every sibling connection must re-authenticate (LOGOFF/LOGON, same token) before next use. Added a **per-server auth epoch**: bumped on `AuthorizationExpired`, stamped on connections when built/re-authed; `ensure_identity` force-re-auths stale-epoch connections (`Connection#authenticate` gains `force:`). Per-address on routing so a reader's expiry doesn't re-auth the writer pool.
4. **Bolt 5.0 routed worker** — same stale-connection discard-and-rebuild for `neo4j://`.
5. **Baseline fold** — 8 NEW passes, 0 regressions.

## Verified locally (boltstub)
- `test_basic_auth_manager` (5x0+5x1): **OK**
- `test_bearer_auth_manager` (5x0+5x1): **OK**
- `test_auth_token_manager` `test_error_on_*`, `test_static_*`: **OK**
- `TestNoRoutingAuthorization` AuthorizationExpired ×2: **OK** (bonus — also completes AuthorizationExpiredTreatment)

## Known remaining (out of scope — routing/home-DB)
`test_dynamic_auth_manager routing=True` (5x0+5x1): expects a `ROUTE` per session (home-DB re-discovery without `Optimization:HomeDatabaseCache`); MRI caches the routing table. Belongs with the routing/home-DB work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of authorization expiration events across pooled connections using per-server generation tracking
  * Enhanced support for session-scoped authentication credentials, including safer security-exception processing
  * Strengthened re-authentication flows and connection rebuilds when in-place re-auth isn’t supported

* **Tests**
  * Updated MRI stub baseline test identifiers to cover the expanded authorization scenarios


klobuczek:
Thanks @copilot-pull-request-reviewer and CodeRabbit — all three findings were valid and are fixed in 8a929df:

1. **Session-scoped auth suppressed the epoch invalidation** (the regression). The provider handler was gated entirely on `!session_scoped`, so a session-scoped `AuthorizationExpired` stopped bumping the per-server epoch and sibling pooled connections wouldn't re-auth. Fixed by splitting the concerns: `notify_security_exception` now **always** calls the handler (so the epoch bump always runs) and passes `session_scoped`; the handler skips only the auth-token-**manager** notification + retryable upgrade for a per-session identity, keeping `handle_security_exception_count` at 0.
2. **Epoch thread-safety.** Direct uses a `Mutex` + an `auth_epoch` reader; routing reads `@auth_epochs` under `@refresh_lock` via `auth_epoch_for`. Each acquire/discovery snapshots the epoch once and uses that single value for both the staleness check and the connection's stamp, so a concurrent bump can't split them.
3. **Bolt 5.0 ROUTE discovery** now runs the same stale-connection discard/retry as the worker path before `conn.route`, so discovery can't ROUTE under a rotated/expired identity.

Re-validated locally (boltstub): `test_basic_auth_manager`, `test_bearer_auth_manager`, the session-auth `test_error_on_*`, and the `AuthorizationExpired` tests all still pass; MRI unit specs green. CI re-running.

PR #384, opened 2026-06-15
mri(routing): close the routing stub-conformance gap

klobuczek:
Closes most of the routing (`neo4j://`) stub-conformance gap, diagnosed via local boltstub runs (and the JRuby/Java reference where behavior was ambiguous). `test_routing_v3` went from **85 failures to ~5**; v5x0 from 6 to 2.

## Root causes fixed
- **Bolt 3.0 HELLO routing context** — the 3.0 HELLO dropped `routing`; testkit expects java/ruby drivers to send it. Failed nearly the whole v3 suite.
- **Bolt 3.0 procedure bookmarks** — the `getRoutingTable` procedure RUN must not carry session bookmarks (bookmark-aware routing is a 4.3+ ROUTE-message feature).
- **Custom + domain-name resolver** — custom address resolver expands the seed into initial routers in the load balancer; domain-name resolver applied at connect (threaded via `getDomainNameResolver`).
- **`max_transaction_retry_time` default 2s → 30s** (match Java) — cluster failover takes seconds; the managed-tx retry must keep rediscovering. Fixes write-retry-on-leader-change/shutdown and resolver-rediscovery.
- **Home-database resolution** — use the resolved `db` from the ROUTE reply (Bolt 4.4+/5.x) on RUN/BEGIN.
- **`supports_multi_db?`** probes the negotiated protocol instead of returning a constant (false on 3.0).

## Known-deferred (separate concern, not this PR)
- `drop_connections_failing_liveness_check` — needs `driver.metrics` (connection-pool metrics; unimplemented on MRI).
- `successfully_acquire_rt_when_router_ip_changes` — needs dead-connection detection on pool-pop + is linux/2-IP gated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of `db` selection across auto-commit, managed transactions, and explicit transaction starts, including correct home-database handling when none is specified.
  * Fixed routing and bookmark payload behavior for Bolt 3 vs Bolt 4 connection routing.
  * Increased default managed transaction retry time to improve resilience.
  * Improved multi-database capability detection by probing protocol support at runtime.
* **Improvements**
  * Enhanced routing-table behavior with resolver-aware hostname expansion and seed ordering.
  * Added direct-path behavior to prevent home-database resolution and enabled optional domain-name resolver support.
* **Tests**
  * Updated stub testkit baseline to cover expanded session/transaction and routing parameter behaviors.


klobuczek:
Thanks @copilot-pull-request-reviewer — all three addressed in 18a6ce7:

1. **Double bookmark snapshot** (`session.rb`): `operation_database` now takes the caller's snapshot instead of computing its own. `run` computes it once and passes it to both the db resolution and the wire `bookmarks`; for explicit/managed transactions the home-db resolution moved into `open_transaction`, which already computes the BEGIN snapshot — so resolution and BEGIN share exactly one snapshot (no double BookmarkManager call, no divergent snapshots). Verified with the bookmark-tx stub tests.
2. **Stale `driver_factory` header comment**: updated to describe `getDomainNameResolver` (returns nil by default, threaded into options) rather than the old raising `get_domain_name_resolver`.
3. **Stale `resolved_addresses` comment**: updated to the domain-name-resolver (hostname→IPs) contract; noted the custom address resolver is the LoadBalancer's routing-only concern.

MRI specs green; default-db, impersonation, and bookmark-tx stub tests pass. CI re-running.

PR #409, opened 2026-07-11
fix(mri): resolve home database per session (home-db + routing-liveness parity)

klobuczek:
## Problem

A routing session with no explicit database (home-db) must re-resolve its home database **per session** — the resolved name depends on the session's context (bookmarks, impersonation, auth). The reference drivers re-`ROUTE` on each acquire when the (SSR-gated) home-db cache doesn't apply.

MRI instead cached the discovery result under the `nil` routing-table key and reused it across sessions, so a second home-db session ran against the **first** session's resolved database (e.g. session 2 sent `RUN` with `db: homedb1` when the router had resolved it to `homedb2`).

## Fix (`Routing::LoadBalancer` + `Session`)

- LoadBalancer never serves the home-db (`nil`) table from cache — it always re-runs discovery — and keys every routing table by its **resolved** database name, never under `nil`.
- `acquire` selects the address by the resolved name returned from discovery.
- `Session` hands the provider the resolved home-db name (memoized in `operation_database`), so there is still exactly **one** `ROUTE` per session and the acquire reuses the table just cached under that name.

## Tests

Locally (MRI backend + Python boltstub), now passing with **no regressions**:

- ✅ All 8 `Test{4x4,5x8}HomeDbWithoutCache` tests (resolve-db-per-session × {session_run, tx_run, tx_func_run} + cache-home-db-despite-new-rt).
- ✅ **Bonus:** `TestLivenessCheckRouting.test_timeout` / `test_timeout_with_re_auth` — same root cause (the router connection is now re-acquired per session, so it gets liveness-probed as the reference expects).
- No get_features flag flipped; `HomeDbWithCache` / `MixedCluster` stay correctly **skipped** (`OPT_HOME_DB_CACHE` unimplemented).
- v4x4 + v5x0 routing suites: only pre-existing/environmental errors (`driver.metrics` unimplemented; a macOS-only IPv6 `fe80::` `getaddrinfo`), confirmed to fail identically on clean `HEAD`.

The 10 new-passes are folded into `testkit-stub-baseline-mri.txt`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved home-database connection routing by using the resolved database consistently for server selection and connection creation.
  - Updated routing-table refresh behavior to cache and return results under the resolved database name.
  - Improved database resolution for connections created from session operations, aligning routing with the operation’s bookmark set.
  - Fixed liveness-check routing for timeout and re-authentication scenarios.
- **Tests**
  - Expanded stub baseline coverage for home-database caching/resolution and liveness-check routing across supported configurations.


klobuczek:
Pushed `e2e32c9` addressing the CI regression and both review comments:

**CI regression (v3/v4.2/v4.3 routing + TestAuthorizationV3)** — root cause was the extra `ensure_routing_table_is_fresh` change that *always* re-routed when `database == nil`. That broke default-database routing on the procedure/pre-4.4 protocols, which legitimately cache the `nil`-keyed table. It turns out that change was **redundant**: keying tables by their **resolved** name (`apply_routing_table`) is sufficient on its own —
- v3/default-db reply carries no `db` → the table stays under `nil` and is cached as before;
- a 4.4+ home-db reply carries the resolved name → `nil` is never populated, so each session re-resolves naturally.

So I reverted the `nil` bypass. This also resolves **@coderabbitai**'s "pre-4.4 home-db acquires still bypass the cache" comment.

**@Copilot** (session.rb) — `acquire_connection` now snapshots `current_bookmarks_for_extra` once and reuses it for both `database:` and `bookmarks:`, since it records `@bookmarks_used_on_begin` and reads the mutable BookmarkManager.

Re-verified locally: the ~40 regressed RoutingV3/V4x2/V4x3 + auth-v3 tests pass again, and the 10 target tests (8 `*HomeDbWithoutCache` + 2 `TestLivenessCheckRouting`) still pass. MRI unit specs green.

PR #410, opened 2026-07-12
fix(mri): verify_connectivity RESET probe — Feature:API:Driver.VerifyConnectivity to green

klobuczek:
## Feature: `Feature:API:Driver.VerifyConnectivity` (MRI-advertised, `'r'`)

11 of its 12 tests passed; the last one, `test_direct_from_pool`, failed. This PR takes the feature fully green.

## Problem

`Direct::ConnectionProvider#verify_connectivity` was `release(acquire)` — it acquired a connection and immediately returned it to the pool **without exercising it on the wire**. A reused (pooled) connection was therefore never probed. testkit's `test_direct_from_pool` asserts `verify_connectivity` sends a RESET on the pooled connection; MRI sent none (`0 == 0`).

## Fix

Send a `RESET` on the acquired connection before releasing it — matching Java's `verifyConnectivity` and the routing `LoadBalancer#verify_connectivity` (which already does this).

## Verification (local, MRI backend + Python boltstub)

- `tests.stub.connectivity_check.test_verify_connectivity`: **24 run / 12 skipped / 0 fail** (the skips are the separate, unadvertised `API:Driver:GetServerInfo`).
- `test_direct` / `test_direct_no_server` / `test_direct_raises_error` still pass (`hello_only.script` has `AUTO RESET`; the MinimalResets-only 0-RESET assertion doesn't apply to MRI).
- MRI unit specs: 61, 0F.

New-pass `test_direct_from_pool` folded into `testkit-stub-baseline-mri.txt`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connectivity verification by actively checking a pooled connection before returning it to the pool.
  * Increased reliability when validating database connectivity, particularly for MRI environments.


klobuczek:
Addressed in `aa0da2a`. `Bolt::Connection#reset!` intentionally swallows `StandardError` for recovery paths, so I added a `reset!(propagate:)` option:

- default `propagate: false` — unchanged swallow behavior for recovery paths;
- `verify_connectivity` passes `propagate: true`, so a failed RESET probe raises instead of reporting false success, and the dead connection is **discarded** (not returned to the pool).

Applied to both the direct and routing `verify_connectivity` paths (same pre-existing pattern). Healthy path unchanged — the full `test_verify_connectivity` suite stays green (24 run / 12 skipped / 0 fail), including the failure-path tests `test_direct_raises_error` and `test_routing_raises_error`.

PR #411, opened 2026-07-12
fix(mri): acquisition timeout bounds the TCP connect — Feature:API:ConnectionAcquisitionTimeout to green

klobuczek:
## Feature: `Feature:API:ConnectionAcquisitionTimeout` (MRI-advertised, `'jar'`)

The last failing test of this feature — `test_should_fail_when_acquisition_timeout_is_reached_first` — hung instead of raising. This PR takes the feature green.

## Problem

`Bolt::Connection#open_socket` bounded the TCP connect only by `connection_timeout`. The test points the driver at a non-routable address (`bolt://10.255.255.255`) with `acquisition_timeout = 2s` but `connection_timeout = 720s`, and expects the **acquisition** timeout to fire first. MRI instead waited on the connection timeout, so the connect hung (harness SIGINT) rather than raising a `DriverError` at 2s.

The `acquisition_deadline` was already applied to the handshake reads — just not to the connect itself.

## Fix

Bound the TCP connect by `min(connection_timeout, connection_acquisition_timeout)`. No-op when `acquisition_timeout` is unset or larger than `connection_timeout` — so the sibling `test_should_fail_when_connection_timeout_is_reached_first` (connection_timeout is the smaller) is unchanged.

## Verification (local, MRI backend + Python boltstub)

- Full `test_connection_acquisition_timeout_ms` class: **10 run / 0 fail** (was 2 errors). Wall-clock dropped 89s → 32s as the hangs are gone.
- MRI unit specs: 61, 0F.

New-pass folded into `testkit-stub-baseline-mri.txt`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Connection attempts now honor one overall acquisition timeout across all address retries and handshake operations.
  * TCP connection timeouts are capped by the remaining acquisition time, preventing retries from exceeding the configured limit.
  * Steady-state read behavior remains unchanged after a successful connection.


klobuczek:
Addressed in the latest commit. Now a single monotonic acquisition deadline is computed once before iterating resolved addresses (stored in `@read_deadline`, which already bounds the handshake/HELLO reads); each `open_socket` uses only the **remaining** budget (`remaining_read_budget`, still capped by `connection_timeout`), and the handshake reuses the same deadline. Retries across addresses can no longer collectively exceed the acquisition timeout. Full class still 10 run / 0 fail; unit specs 61/0.

PR #412, opened 2026-07-12
feat(mri): connection-pool metrics — Feature:API:Liveness.Check drop_connections

klobuczek:
## Feature: connection-pool metrics (the documented MRI TODO)

`test_should_drop_connections_failing_liveness_check` (×5 Bolt versions, gated on `Feature:API:Liveness.Check`) errored on MRI with `undefined method 'metrics'` — `Driver#pool_metrics` raised `NotImplementedError` and no pool-metrics infra existed. This implements it.

- **`Bolt::Pool`** tracks two mutex-guarded counters: `@created` (live connections built) and `@leased` (checked out) → `idle = created - leased`, `in_use = leased`. Bumped in the create block / `prepare` / `push` / `discard` / `discard_on_pop`. Pure bookkeeping — no wire effect.
- **`Direct::ConnectionProvider`** and **`Routing::LoadBalancer`** expose `connection_pool_metrics` (one entry per server-address pool).
- **`Internal::Metrics`** wraps the provider; **`Driver#metrics`** returns it, matching testkit's `GetConnectionPoolMetrics` (`.address`/`.in_use`/`.idle`).

MRI owns its pool, so this is viable where JRuby's is blocked (shaded-jar ABI clash — jruby's `drop_connections` stays skipped).

## Note on validation

The metrics API is confirmed working locally (the `undefined method` error is gone). The `drop_connections` test additionally exercises the liveness-drop path; per the design memo, **metrics is the sole documented MRI gap** for this test. Local repro of the drop itself is unfaithful (the python boltstub's `A: RESET` handling differs from CI's rust boltstub — a pattern seen repeatedly), so I'm relying on CI to confirm the 5 `drop_connections` tests go green, then I'll fold them into the baseline.

MRI unit specs: 61, 0F.

## Summary by CodeRabbit

* **New Features**
  * Added a public `metrics` API on the driver for connection-pool metrics.
  * Exposed per-server pool snapshots with active (`in_use`) and idle connection counts for direct and routed drivers.
* **Bug Fixes**
  * Reworked connection-pool metrics accounting to keep counts accurate across checkout, return, liveness failures, discard, and shutdown.
* **Tests**
  * Updated pool metrics specs and adjusted MRI testkit baseline to cover new liveness-check and timeout scenarios.


klobuczek:
Addressed the review in the latest commit:

- **Atomicity (Copilot + CodeRabbit):** `@created`/`@leased` now move under a single lock per update (`bump(created:, leased:)`), and reads go through `metrics_snapshot -> [in_use, idle]` (one lock) — so a concurrent read can't observe a half-applied discard / negative idle. The providers use the snapshot.
- **Shutdown (CodeRabbit):** `Bolt::Pool#shutdown` now zeroes the counters, so a post-close metrics request reports nothing live.
- **`id` (Copilot / DECISIONS.md):** `ConnectionPoolMetrics#id` returns `"host:port"` for the forward-compatible duck-typed contract, keeping `#address` for the current handler.
- **Tests (Copilot):** `pool_spec` now covers `in_use`/`idle` across pop / push / discard / replace-on-pop / shutdown.

MRI unit specs 66/0.

Separately — I confirmed against the **rust** boltstub locally (same as CI) that the 5 `drop_connections` tests need more than metrics: MRI doesn't send a RESET on pool-return, so the liveness probe is absorbed by the script's `A: RESET` and the stale connections aren't dropped. Working that next; it may land here or as a follow-up depending on blast radius.

PR #413, opened 2026-07-13
fix(mri): retry classification + write-failover reuse — Feature:Bolt:4.3 retry to green

klobuczek:
## Feature: `Feature:Bolt:4.3` — the retry failures

Four failing retry tests, two distinct fixes.

### Non-retryable reclassification (`test_should_not_retry_non_retryable_tx_failures` ×2)
`Neo.TransientError.Transaction.{Terminated,LockClientStopped}` are `TransientError` by category but non-retryable. The driver now rewrites them to their `ClientError` code, so the managed-tx executor treats them as non-retryable `ClientException`s and surfaces the rewritten code (`bolt/message/failure.rb`).

### Write-failover reuse (`test_retry_NotALeader`, `test_retry_ForbiddenOnReadOnlyDatabase`)
On a `NotALeader` / `ForbiddenOnReadOnlyDatabase` write failure, the node is no longer the leader but the **connection is fine**. The driver now forgets it as a writer (forcing a re-route) but **does not discard it** — the RESET-on-release added in #412 clears the FAILED state so the retry reuses the connection, matching Java's `onWriteFailure` and testkit's `retry_with_fail_after_pull` script (which retries on the same connection). Previously MRI discarded the connection mid-script (→ `GOODBYE` where the script expected the retry `BEGIN`) and the re-route dead-ended on `SessionExpired: No write servers available`.

## Verification (rust boltstub = CI-faithful)

- All 4 retry tests pass.
- Full retry suites green (`test_retry` 6/0, `test_retry_clustering` 8/0).
- Routing v5x0 / v4x4 / v4x3 sweeps: only the pre-existing environmental `router_ip_changes` (macOS IPv6) fails — no new regressions from the write-failover change.
- MRI unit specs 66/0.

The 4 tests are folded into `testkit-stub-baseline-mri.txt`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of transaction termination and lock-related errors so they are reported with the correct client error classifications.
  * Enhanced write-routing failure recovery by retaining reusable connections while refreshing the affected writer route.
  * Preserved the original error code when reporting expired sessions after write failures.

* **Tests**
  * Updated integration and stub test baselines to cover transaction timeouts and additional retry scenarios.


klobuczek:
Fixed in the latest commit — updated the `classify_failure` method-level comment to say the NotALeader/Forbidden write-failover branch forgets the writer (forces a re-route) but keeps the connection (RESET on release makes it reusable), matching the code.

PR #428, opened 2026-07-15
feat(mri): surface Bolt 6.0 UnsupportedType [Feature:API:Type.UnsupportedType]

klobuczek:
Next `j`-but-missing-`r` feature. Reuses the deferred-value groundwork idea from #427 (Temporal), but here the value is *returned* (serialises to `CypherUnsupportedType`), not raised.

## What
Bolt 6.0 can push a forward-compat marker struct (signature `0x3F`) for a value type newer than the driver understands. `V6` already had a `0x3F` handler that hydrated it to a plain hash "so it doesn't crash" — this wires it to a real public type.

- **New `Types::UnsupportedType`** — `name`, `min_protocol_version` (`"major.minor"`), optional `message`. Mirrors the Java driver's `UnsupportedType`.
- **`V6`'s `0x3F` handler** builds it. The struct's 4th field is an *extra map* whose `:message` key carries the server note (empty map ⇒ no message).
- **`conversion.rb`** serialises the MRI type to `CypherUnsupportedType` (next to the existing JRuby/Java case).
- Advertise `Feature:API:Type.UnsupportedType` (`'ja'` → `'jar'`).

## Validation (rust boltstub)
- `test_unsupported_type` + `test_unsupported_type_in_list` **2/0** (with/without message, bare and nested in a list).
- Full `datatypes` 31/0; MRI unit specs 69/0.

## Summary by CodeRabbit

* **New Features**
  * Added support for representing and inspecting unsupported Bolt types (name, minimum protocol version, and optional server message) across both MRI and JRuby.
* **Bug Fixes**
  * Prevented unsupported values from being packed as query parameters.
  * Improved unsupported type hydration, including safer handling of VECTOR.
  * Updated conversion behavior in the testkit backend to correctly map unsupported types.
  * Corrected feature detection for unsupported type support.
* **Documentation**
  * Clarified Bolt 6 hydration behavior for unsupported result records and VECTOR handling.

### DIFF
--- a/testkit-backend/conversion.rb
+++ b/testkit-backend/conversion.rb
@@ -53,14 +53,15 @@ module TestkitBackend
                          startNodeElementId: to_testkit(object.start_node_element_id),
                          endNodeElementId: to_testkit(object.end_node_element_id),
                          type: to_testkit(object.type), props: to_testkit(object.properties))
-          when ->(o) { defined?(Java::OrgNeo4jDriverTypes::UnsupportedType) &&
-                       Java::OrgNeo4jDriverTypes::UnsupportedType === o }
-            # A value of a type newer than this driver understands; the Java
-            # driver keeps its name + the bolt version that introduced it.
-            # JRuby-only type, so guard the Java constant (absent on MRI).
+          when Neo4j::Driver::Types::UnsupportedType
+            # A value whose type is newer than the driver understands (name +
+            # the protocol version that introduced it + optional server note).
+            # One case for both flavours: MRI's pure-Ruby type and, aliased in
+            # lib/jruby, the Java type — whose #message is unwrapped to a plain
+            # nilable there (ext/unsupported_type.rb) so both APIs match.
             named_entity('CypherUnsupportedType', name: object.name,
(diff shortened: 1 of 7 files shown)


testkit-backend/conversion.rb
```diff
@@ -53,6 +53,13 @@ def to_testkit(object)
...
+          when Neo4j::Driver::Types::UnsupportedType
+            # MRI (pure-Ruby) UnsupportedType — a Bolt 6.0+ struct newer than
+            # the driver understands (name + the protocol version that added it
+            # + optional server message).
+            named_entity('CypherUnsupportedType', name: object.name,
+                         minimumProtocol: object.min_protocol_version,
+                         message: object.message)
           when ->(o) { defined?(Java::OrgNeo4jDriverTypes::UnsupportedType) &&
                        Java::OrgNeo4jDriverTypes::UnsupportedType === o }
             # A value of a type newer than this driver understands; the Java
```
klobuczek: We need to collapse those 2 cases to 1. Add the type alias in jruby, similar to Node, Path etc.

PR #432, opened 2026-07-15
feat(mri): pipeline BEGIN+RUN+PULL in execute_query [Optimization:ExecuteQueryPipelining]

klobuczek:
Closes the MRI gap on `Optimization:ExecuteQueryPipelining` (JRuby wraps the Java driver, which already pipelines; MRI now does too — `ja` → `jar`).

## What

`Driver#execute_query` runs a managed auto-commit transaction. Previously BEGIN was a discrete round-trip (sent, then its SUCCESS read) before RUN+PULL could be flushed. Testkit's `begin_pipeline.script` (Bolt 5.3) instead expects all three client messages before any server reply:

```
C: BEGIN
   RUN
   PULL
S: SUCCESS {}          # begin
   SUCCESS {"fields": ["n"]}   # run
   RECORD [1]
   SUCCESS {"has_more": false} # pull
C: COMMIT
S: SUCCESS {}
```

This defers reading BEGIN's reply so the first RUN+PULL hit the wire before the driver blocks — saving two round-trips.

## How

- `Transaction` gains a `pipelined:` flag. When set, `#initialize` flushes BEGIN but skips its reply; `#run` drains it (`#ack_begin!`) right after flushing RUN+PULL, then reads RUN's reply. **Every non-pipelined transaction reads BEGIN's reply eagerly in `#initialize` exactly as before** — those paths (explicit tx, user-called `execute_read`/`write`, and all their BEGIN-failure handling) are byte-identical.
- `#ack_begin!` is idempotent: it reads the deferred BEGIN + the telemetry SUCCESS pipelined ahead of it, and fires the telemetry-ack callback. `rollback` drains a deferred BEGIN defensively for a pipelined tx whose query never ran (e.g. local validation failure), keeping the reply queue aligned.
- Only `execute_query`'s internal session pipelines — it reports telemetry api `3` (`DRIVER_EXECUTE_QUERY`), a stable session-scoped signal, so `open_transaction` sets `pipelined:` on that alone.

## Testing

rust boltstub (`TEST_RUSTY_STUB=true`, MRI backend):
- `optimizations::test_execute_query_pipelines_begin` — passes (the target)
- Regression: `driver_execute_query` 15/0, `tx_run` 19/0, `tx_begin_parameters` 16/0, `iteration` 56/0, `errors` 8/0, `disconnects` 12/0, full `optimizations` 8/0
- MRI unit specs — 69/0

## Summary by CodeRabbit

* **Performance**
  * Improved `executeQuery` transaction handling by pipelining transaction start, query execution, and result retrieval for faster processing.

* **Bug Fixes**
  * Improved rollback handling for pipelined transactions that are closed before query execution begins.

* **Compatibility**
  * Updated feature detection so supported JRuby and MRI drivers correctly recognize execute-query pipelining.

### DIFF
--- a/lib/mri/neo4j/driver/session.rb
+++ b/lib/mri/neo4j/driver/session.rb
@@ -485,6 +485,9 @@ module Neo4j
         tx_options = tx_options.merge(database: operation_database(bookmarks)).compact
         Transaction.new(connection, self, bookmarks, tx_options, telemetry_api: telemetry_api,
                         telemetry_ack: telemetry_ack,
+                        # executeQuery's session reports telemetry api 3 (DRIVER_EXECUTE_QUERY);
+                        # only that path pipelines BEGIN + RUN + PULL (Optimization:ExecuteQueryPipelining).
+                        pipelined: @options[:telemetry_api] == 3,
                         on_release: -> { @connection_provider.release(connection) })
       end
     end
--- a/lib/mri/neo4j/driver/transaction.rb
+++ b/lib/mri/neo4j/driver/transaction.rb
@@ -6,10 +6,14 @@ module Neo4j
     class Transaction
       attr_reader :connection
 
-      def initialize(connection, session, bookmarks = [], options = {}, telemetry_api: nil, telemetry_ack: nil, on_release: nil)
+      def initialize(connection, session, bookmarks = [], options = {}, telemetry_api: nil, telemetry_ack: nil,
+                     pipelined: false, on_release: nil)
         @connection = connection
         @session = session
         @options = options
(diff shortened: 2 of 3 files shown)


klobuczek:
Good catch — fixed in the latest commit. Replaced the blocking `ack_begin!` drain in `rollback` with `rollback_via_reset` for the pipelined-never-ran case, exactly as you suggested: RESET rolls the tx back and `drain_quiesced` drains any pending reply (the un-acked BEGIN + RESET's own) with no blocking read, so a server withholding BEGIN's SUCCESS can't deadlock. There are no open results on that path, so nothing else needs draining. Target test + driver_execute_query/tx_run/errors stay green.

PR #435, opened 2026-07-16
feat(mri): retry idempotent auto-commit RUN once [Feature:IdempotentRetries]

klobuczek:
Closes the MRI gap on `Feature:IdempotentRetries` (`ja` → `jar`).

## What

An auto-commit `session.run` whose **RUN reply** is a server FAILURE flagged `_idempotent` in its diagnostic record is retried **once**: `RESET` clears the FAILED state (and drains the pipelined PULL's `IGNORED`), then RUN+PULL are re-sent — **without** re-sending `TELEMETRY`. A second failure (idempotent or not) is raised, as is a non-idempotent first failure.

Scope guards (verified against the stub scripts, which flag `_idempotent` on telemetry/pull failures too):
- A `stage` marker keeps a **TELEMETRY**-reply failure and send/flush transport errors out of the retry path.
- A **PULL**/stream failure surfaces during iteration as before (the RUN already succeeded).
- **Explicit** (`begin_transaction`) and **managed** transactions are unaffected — this is the auto-commit path only.

Config (tri-state, matches Java): session-level `AutoCommitRetriesMode` wins; otherwise the driver-level `auto_commit_retries_disabled` default applies (defaulting to enabled). Both were already plumbed from the testkit-backend and flow via `driver.session`'s option merge — no new wiring.

## Testing

rust boltstub (MRI backend):
- `tests.stub.idempotent_retries` **8/8** — retry-then-success, telemetry-not-resent, two-errors-give-up, error-on-pull/telemetry not retried, the full driver×session config matrix, explicit-tx no-retry
- Regression: `iteration` 56/0, `tx_run` 19/0, `errors` 8/0, `summary` 132/0, `driver_execute_query` 15/0, `disconnects` 12/0; MRI unit specs 69/0

## Summary by CodeRabbit

* **New Features**
  * Auto-commit queries can now be retried once when the server identifies a failure as safe to retry.
  * Retry behavior is controlled through session or driver settings.

* **Bug Fixes**
  * Prevented retries for telemetry, streaming, pull, and transport failures that are not safe to repeat.
  * Improved connection recovery before retrying eligible queries.

### DIFF
--- a/lib/mri/neo4j/driver/auto_commit_retries_mode.rb
+++ b/lib/mri/neo4j/driver/auto_commit_retries_mode.rb
@@ -6,9 +6,9 @@ module Neo4j
     # (Feature:IdempotentRetries) — the MRI counterpart of the Java driver's
     # org.neo4j.driver.AutoCommitRetriesMode enum. Same constant names, MRI's
     # own (symbol) values, mirroring how AccessMode/RoutingControl are defined
-    # here. Defined so the shared testkit-backend can name the mode
-    # flavor-agnostically; the pure-Ruby Bolt 6.0 retry path is not yet
-    # implemented, so MRI does not act on it.
+    # here. Session#run acts on it (together with the driver-level
+    # `auto_commit_retries_disabled` default) to retry an idempotent auto-commit
+    # RUN once — see Session#auto_commit_retries_enabled?.
     module AutoCommitRetriesMode
       ENABLED = :enabled
       DISABLED = :disabled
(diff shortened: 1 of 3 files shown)


klobuczek:
Both addressed in the latest commit:

1. **Telemetry lease leak** — moved the `connection.telemetry(...)` send back inside the release-protected `begin` block (guarded by a `telemetry_sent` once-flag so a retry doesn't resend it). A send failure now hits the `rescue` and releases the lease.
2. **Ineffective `auth_failed` guard** — the RUN failure is now classified (`classify_failure`) *before* the retry decision, so a security failure sets `auth_failed` and the `!connection.auth_failed` guard actually holds. `idempotent_error?` is still evaluated on the original exception, which carries the diagnostic record.

idempotent_retries 8/0, iteration 56/0, errors 8/0, MRI specs 69/0.

PR #436, opened 2026-07-17
feat(mri): optimistic home-database cache [Optimization:HomeDatabaseCache]

klobuczek:
Closes the MRI gap on `Optimization:HomeDatabaseCache` (`ja` → `jar`).

## What

A routed driver can skip home-database discovery by reusing an identity's most recently resolved home db — safe only when **every open connection advertises server-side routing** (Bolt 5.8 `ssr.enabled`), so the server transparently re-routes a wrong guess. Mirrors the Java/Python drivers.

## How

- **SSR detection & gating**: `Connection#ssr_enabled?` parses the `ssr.enabled` HELLO hint; a single `#on_close` fires across all teardown paths. `LoadBalancer` maintains a pool-wide SSR tally (`track_ssr_open/close`) and only consults the cache when all open connections are SSR.
- **`Internal::HomeDbCache`**: bounded (10k) LRU keyed by `imp_user`, else the per-session auth token (driver-level auth → one shared key). Matches Java — a basic principal is **not** folded into `imp_user`, so `HomeDbCacheBasicPrincipalIsImpersonatedUser` stays unadvertised (its test correctly skips).
- **Resolution semantics** (`Session#acquire_for_operation`) — the subtle part, verified against the stub scripts:
  - A guess is used **only when its routing table is already fresh** → acquire against it and send `db=null` so the server resolves the real home db (cache updates from the reply's `db`).
  - A **stale** guess or a **non-SSR** connection (mixed cluster) → fall back to discovery (`ROUTE db=null`) and **pin** the resolved name.
  - The session pins its home db after first resolution so a re-resolving server can't move it mid-session; **direct drivers never pin**.
- **Shared acquisition deadline**: the optimistic acquire and its fallback share one budget, threaded `provider#acquire → Pool#pop → Connection#connect` (bounding the handshake too).
- Wired for **session.run**, **explicit/managed tx** (`Transaction#on_begin`), and **executeQuery**. `AuthTokens.custom` now tolerates nil `parameters` (a backend-glue crash the un-skipped tests exposed).

## Testing

rust boltstub (MRI backend): **`tests.stub.homedb` 37/37** — `TestHomeDbWithCache`/`Tx`/`ExecuteQuery` + mixed-cluster fallback (incl. `acquisition_timeout_during_fallback`); 3 sub-feature skips.
No regressions: `tx_run`, `iteration`, `disconnects`, `errors`, `summary`, `driver_execute_query`, `idempotent_retries`, `optimizations`, `configuration_hints`, routing (pool-size/liveness); MRI unit specs 69/0. (The one routing `router_ip_changes` error is the known link-local-IPv6 local flake.)

## Summary by CodeRabbit

* **New Features**
  * Improved home database selection with driver-level LRU caching, per-session home database pinning, and cache learning.
  * Added optional connection acquisition deadlines end-to-end (including routing/pool/direct providers).
  * Added SSR capability detection and tracking for pooled routing connections; SSR is explicitly disabled for direct connections.
* **Bug Fixes**
  * Improved authentication token parameter handling to avoid assigning nil/empty parameters.
  * Improved transaction BEGIN callback delivery.
  * Enhanced connection teardown to invoke the close callback exactly once.
* **Tests / Chores**
  * Updated feature-tag advertisement for the home database cache optimization.

### DIFF
--- a/lib/mri/neo4j/driver/auth_tokens.rb
+++ b/lib/mri/neo4j/driver/auth_tokens.rb
@@ -32,7 +32,7 @@ module Neo4j
         def custom(principal, credentials, realm, scheme, parameters = {})
           token = { scheme: scheme, principal: principal, credentials: credentials }
           token[:realm] = realm if realm
-          token[:parameters] = parameters unless parameters.empty?
+          token[:parameters] = parameters unless parameters.nil? || parameters.empty?
           token
         end
       end
--- a/lib/mri/neo4j/driver/bolt/connection.rb
+++ b/lib/mri/neo4j/driver/bolt/connection.rb
@@ -81,6 +81,10 @@ module Neo4j
           # LOGOFF/LOGON replies pipelined ahead of the next operation and not yet
           # consumed (Optimization:AuthPipelining).
           @pending_auth_acks = 0
+          # Guards the one-shot #on_close swap: close (caller thread) and
+          # mark_closed_broken (reader thread) can race, and firing the callback
+          # twice would double-decrement the routing SSR tally.
+          @on_close_mutex = Mutex.new
           @recv_timeout = nil # server's connection.recv_timeout_seconds hint
           @read_deadline = nil # monotonic bound for acquisition-phase reads
           # Writes go behind a mutex: the reader writes watermark follow-up
@@ -111,7 +115,7 @@ module Neo4j
           @auth_epoch = 0
         end
 
-        def connect
+        def connect(deadline: nil)
           last_error = nil
           # One monotonic acquisition deadline for the whole connect, shared by
           # every resolved-address attempt AND the handshake/HELLO reads: a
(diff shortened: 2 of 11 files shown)


klobuczek:
Pushed a rebase onto current main + the review fixes:

- **Regressions** (auth_token_manager / auth-scheme tests): these were against the branch's **stale base** (predating the merged auth PRs). Rebased onto current main; the full `tests.stub.authorization` suite is now **168/168** locally, and homedb 37/37, MRI specs 69/0.
- **Dead code** (@Copilot): deleted `Session#acquire_connection` and `Session#operation_database` — both are superseded by `acquire_for_operation` and had no remaining callers.
- **fire_on_close race** (@coderabbitai): the read-then-clear of `@on_close` could fire twice (close vs the reader's `mark_closed_broken`) and double-decrement the SSR tally. Now serialized with a dedicated `@on_close_mutex` (callback invoked outside the lock, since it re-enters the load balancer's SSR mutex).

klobuczek:
Fixed the two follow-ups:

- **Bookmark-manager regression** (test_should_enrich_bookmarks_with_bookmark_supplier_result / test_should_not_change_bmm_state): my `acquire_for_operation` reused the RUN/BEGIN bookmark snapshot for discovery, so a BookmarkManager supplier was consulted once per op instead of per-acquire (2 vs the expected 4/5). Restored the separate snapshot — `acquire_for_operation` now takes its own for discovery/acquire, and RUN/BEGIN keep theirs (BEGIN's taken last so `@bookmarks_used_on_begin` stays correct). test_bookmark_manager 16/16, homedb 37/37, tx_run/iteration green.
- **on_close= race** (@coderabbitai): replaced the `attr_accessor` with a setter that fires the callback immediately if the connection already tore down between build and assignment, so the SSR tally can't miss a decrement.

PR #439, opened 2026-07-19
feat(mri): mutual-TLS client certificates [Feature:API:SSLClientCertificate]

klobuczek:
## Summary

Implements the `ClientCertificate` / `ClientCertificateManager` API for the **MRI** flavour so the pure-Ruby Bolt driver can present a client certificate during the TLS handshake (mutual TLS). This closes the `Feature:API:SSLClientCertificate` testkit gap — JRuby already advertised it via the Java driver; MRI now advertises it too.

## What's here

- **`ClientCertificate`** value object + **`ClientCertificates.of(certfile, keyfile, password = nil)`** factory.
- **`ClientCertificateManager`** — block-based; caches the last non-nil certificate so a `nil` reply means "keep the current one" (matches Java semantics). **`ClientCertificateManagers.rotating(cert)`** helper.
- **`DriverFactory#new_instance`** threads the manager into the per-connection options (previously accepted-but-ignored on MRI).
- **`TlsConfig#apply_client_certificate`** polls the manager on each connect and sets `ctx.cert` / `ctx.key`. `OpenSSL::PKey.read` handles both encrypted (password) and plain PEM keys. Polling per-connect is what makes **rotation** work — the next connection picks up a rotated cert.
- **JRuby flavour**: `ClientCertificates` / `ClientCertificateManagers` delegate to the Java factories (wrapping paths in `java.io.File`), keeping the testkit backend flavour-agnostic.
- **Backend**: flavour-agnostic `Request.client_certificate_from` + `NewClientCertificateProvider` round-trips to the frontend for cert updates (blocking, like `NewAuthTokenManager`).
- Advertise `Feature:API:SSLClientCertificate` for both flavours.

## Validation

Ran `tests.tls.test_client_certificate` locally against the MRI backend (CRuby) with the test CA trusted — **all 5 cases pass**:

- `neo4j+s` / `bolt+s` with client certificate present
- `neo4j+ssc` / `bolt+ssc` (trust-all) with and without client certificate
- `+s` / `+ssc` with certificate not present
- **client rotation** (`TestClientCertificateRotation`)

CI `testkit-tls (mri)` will exercise the `+s` schemes in the Docker environment where the test CA is imported into the trust store.

## Summary by CodeRabbit

- **New Features**
  - Added mutual TLS client-certificate support for MRI and JRuby.
  - Added APIs for creating and rotating client certificates.
  - Added support for encrypted private keys.
  - Enabled client certificates in driver connections and advertised SSL client-certificate support across platforms.

- **Bug Fixes**
  - Fixed client-certificate settings not being applied to MRI driver connections.

### DIFF
--- a/lib/mri/neo4j/driver/bolt/tls_config.rb
+++ b/lib/mri/neo4j/driver/bolt/tls_config.rb
@@ -36,6 +36,7 @@ module Neo4j
           ctx = OpenSSL::SSL::SSLContext.new
           ctx.min_version = OpenSSL::SSL::TLS1_2_VERSION
           apply_trust(ctx)
+          apply_client_certificate(ctx)
           ctx
         end
 
@@ -50,6 +51,22 @@ module Neo4j
         end
 
         private
+
+        # Mutual TLS (Feature:API:SSLClientCertificate): present the client
+        # certificate the manager currently holds. Polled per connection so a
+        # rotated certificate takes effect on the next connect; nil manager or
+        # nil certificate leaves the context client-cert-free.
+        def apply_client_certificate(ctx)
+          certificate = @options[:client_certificate_manager]&.get_client_certificate
+          return unless certificate
+
+          ctx.cert = OpenSSL::X509::Certificate.new(File.read(certificate.certfile))
+          # PKey.read handles both encrypted (password used) and unencrypted
+          # (password ignored) PEM keys. Pass "" rather than nil for a missing
+          # password so an encrypted key raises PKeyError instead of blocking
+          # the thread on OpenSSL's interactive stdin passphrase prompt.
+          ctx.key = OpenSSL::PKey.read(File.read(certificate.keyfile), certificate.password || '')
+        end
 
         def apply_trust(ctx)
           case resolved_strategy
--- a/lib/mri/neo4j/driver/client_certificate_manager.rb
+++ b/lib/mri/neo4j/driver/client_certificate_manager.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Neo4j
+  module Driver
+    # Supplies the current client certificate for mutual TLS
+    # (Feature:API:SSLClientCertificate). Wrap a block that returns a
+    # ClientCertificate — or nil when nothing changed since the last call, in
+    # which case the previously supplied certificate is kept. Pass an instance
+    # to GraphDatabase.driver via `client_certificate_manager:`; the driver
+    # consults it whenever a connection needs a (possibly rotated) certificate.
+    #
+    # The MRI counterpart of the Java driver's org.neo4j.driver
+    # .ClientCertificateManager (whose get returns null for "no change"); the
+    # last non-nil certificate is retained here so a fresh per-connection TLS
+    # context always presents the current one.
+    class ClientCertificateManager
+      def initialize(&get_client_certificate)
+        @get_client_certificate = get_client_certificate
+        @current = nil
+        @mutex = Mutex.new
+      end
+
+      # The current client certificate. Polls the block; a non-nil result
+      # rotates the retained certificate, a nil result keeps it. Connections
+      # are acquired concurrently, so the poll-and-update is serialized: it
+      # keeps @current consistent and stops a blocking provider (the testkit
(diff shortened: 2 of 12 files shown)


lib/jruby/neo4j/driver/client_certificate_managers.rb
```diff
@@ -0,0 +1,14 @@
...
+    # JRuby flavour: delegate to org.neo4j.driver.ClientCertificateManagers so
+    # the flavour-agnostic testkit backend can call
+    # ClientCertificateManagers.rotating(cert) uniformly.
+    module ClientCertificateManagers
+      def self.rotating(client_certificate)
+        Java::OrgNeo4jDriver::ClientCertificateManagers.rotating(client_certificate)
+      end
+    end
+  end
+end
```
klobuczek: The whole file is unnecessary. This class is already imported.

lib/jruby/neo4j/driver/client_certificates.rb
```diff
@@ -0,0 +1,21 @@
...
+        key = java.io.File.new(keyfile.to_s)
+        if password
+          Java::OrgNeo4jDriver::ClientCertificates.of(cert, key, password)
+        else
+          Java::OrgNeo4jDriver::ClientCertificates.of(cert, key)
+        end
+      end
+    end
+  end
+end
```
klobuczek: Do this in a similar way to GraphDatabase.driver rather than shadowing the already existing ClientCertificates class.

lib/mri/neo4j/driver/client_certificate_manager.rb
```diff
@@ -0,0 +1,31 @@
...
+        @current = nil
+      end
+
+      # The current client certificate. Polls the block; a non-nil result
+      # rotates the retained certificate, a nil result keeps it.
+      def get_client_certificate
+        certificate = @get_client_certificate.call
+        @current = certificate if certificate
+        @current
+      end
```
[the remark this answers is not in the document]
klobuczek: Done — `get_client_certificate` now serializes the poll-and-update under a `Mutex`, so `@current` stays consistent and a blocking provider (the testkit callback) is never entered concurrently.

lib/mri/neo4j/driver/bolt/tls_config.rb
```diff
@@ -51,6 +52,20 @@ def verify_hostname?
...
+        # rotated certificate takes effect on the next connect; nil manager or
+        # nil certificate leaves the context client-cert-free.
+        def apply_client_certificate(ctx)
+          certificate = @options[:client_certificate_manager]&.get_client_certificate
+          return unless certificate
+
+          ctx.cert = OpenSSL::X509::Certificate.new(File.read(certificate.certfile))
+          # PKey.read handles both encrypted (password used) and unencrypted
+          # (password ignored) PEM keys.
+          ctx.key = OpenSSL::PKey.read(File.read(certificate.keyfile), certificate.password)
```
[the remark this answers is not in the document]
klobuczek: Done — `OpenSSL::PKey.read` now receives `certificate.password || ''`. Verified on the fixtures: unencrypted key + '' loads, encrypted key + password loads, and an encrypted key with no password raises `PKeyError` instead of blocking on the stdin passphrase prompt.

lib/jruby/neo4j/driver/client_certificate_managers.rb
klobuczek: Done — deleted the file. `Neo4j::Driver::ClientCertificateManagers` already resolves to the Java class via `include_package 'org.neo4j.driver'`, and its `.rotating(ClientCertificate)` is exactly what the wrapper delegated to, so the backend now calls the Java factory directly.

lib/jruby/neo4j/driver/client_certificates.rb
klobuczek: Done — removed the shadowing module and instead prepend `Ext::ClientCertificates` onto `org.neo4j.driver.ClientCertificates`'s singleton class (like `Ext::GraphDatabase` / `Ext::BookmarkManagers`), overriding `.of` to wrap path strings in `java.io.File` and `super` into the Java factory.

PR #440, opened 2026-07-20
feat(mri): reject scheme+manual security config [Detail:DefaultSecurityConfigValueEquality]

klobuczek:
## Summary

Implements `Detail:DefaultSecurityConfigValueEquality` behaviour for the **MRI** flavour.

A `+s`/`+ssc` URI scheme already fixes the security plan (encrypted; plus trust-all for `+ssc`). Combining it with *manually-configured* encryption/trust settings is a conflict — the MRI driver now raises `ClientException("Scheme <scheme> is not configurable with manual encryption and trust settings")`, mirroring Java's `SecurityPlans.assertSecuritySettingsNotUserConfigured`.

**Value equality** (the feature name): a setting whose value equals the driver default (encryption off; trust = system certificates) is treated as *not configured*, so it is not a conflict — matching Java's `SecuritySettings.isCustomized`. Hence `encryption: false` / no trust strategy on a `+s` scheme is accepted, while `encryption: true` or an explicit trust strategy raises.

## Changes

- **`lib/mri/.../internal/driver_factory.rb`** — `DriverFactory#validate_security_settings` (+ value-equality helper), called from `new_instance`.
- **`spec/mri/.../internal/driver_factory_spec.rb`** — 9 cases across scheme × config × value-equality.
- **`testkit-backend/requests/get_features.rb`** — advertise the feature for MRI (`ja` → `jar`).
- **`testkit/testkit.json`** — bump the pin to latest 6.x (`ec48679` → `840cf46`).

## No testkit change here (by design)

The feature's only testkit test, `tests/tls/test_explicit_options::test_explicit_config_and_scheme_config`, is **deliberately not registered in `tests/tls/suites.py`** upstream (it has never been, since the module was created) and its `else: self.fail("Add expected error type for driver.")` would break drivers without a branch (python, go). Registering it in the shared suite is a testkit-maintainer decision, not ours — so this PR does **not** touch testkit.

Consequence: the conformance test doesn't run in our CI, so the advertisement is inert (nothing runs, nothing fails). But the behaviour is implemented and unit-tested, so ruby passes the moment that test is added to a suite. The small "add `ruby` to the test's driver-name branch" change is parked on a fork branch for a future upstream testkit PR.

## Validation

- `spec/mri/.../driver_factory_spec.rb` — all 9 cases (run in CI against a live Neo4j).
- Behaviour byte-verified against Java 6.1.0's `SecurityPlans.isCustomized`.
- Existing TLS suite unaffected: the `+s` secure-scheme classes construct drivers with empty config, so the new validation never triggers.

> Note: the base bump (`ec48679` → `840cf46`) also pulls upstream's new UUID Bolt 6.1 integration tests, which may surface as new failures in the integration suite (they need a live 6.1 server).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of security settings for encrypted connection schemes.
  * Raised clearer client errors when encrypted URI schemes conflict with manually configured encryption or trust settings.
  * Default and non-conflicting settings for encrypted and unencrypted URIs are accepted.
* **Tests**
  * Added coverage for security configuration combinations across encrypted and standard URI schemes.
* **Chores**
  * Updated testkit feature metadata and refreshed the pinned testkit revision.

### DIFF
--- a/lib/mri/neo4j/driver/internal/driver_factory.rb
+++ b/lib/mri/neo4j/driver/internal/driver_factory.rb
@@ -51,7 +51,8 @@ module Neo4j
         # (rotatable) client certificate. nil (the common case) leaves TLS
         # unchanged.
         def new_instance(uri, auth_token_manager, client_certificate_manager: nil, **config)
-          validate_uri(uri)
+          parsed_uri = validate_uri(uri)
+          validate_security_settings(parsed_uri.scheme, config)
           config = config.merge(client_certificate_manager: client_certificate_manager) if client_certificate_manager
           # The factory is the single place that knows about the non-public
           # extension hooks (the domain-name resolver and the clock). It
@@ -69,7 +70,7 @@ module Neo4j
           # connection / routing / session, never mixed into the user-facing
           # `config`.
           clock = create_clock
-          provider = build_connection_provider(URI(uri), auth_token_manager, config, clock)
+          provider = build_connection_provider(parsed_uri, auth_token_manager, config, clock)
           Driver.new(uri, config, provider, clock: clock)
         end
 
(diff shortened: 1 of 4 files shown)


spec/mri/neo4j/driver/internal/driver_factory_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    context 'with a plain bolt / neo4j scheme' do
+      it 'accepts manual encryption and trust settings' do
+        expect { build('bolt://localhost', encryption: true) }.not_to raise_error
+        expect { build('neo4j://localhost', trust_strategy: { strategy: :trust_all_certificates }) }
+          .not_to raise_error
+      end
+    end
+  end
+end
```
klobuczek: can't those be shared specs?

lib/mri/neo4j/driver/internal/driver_factory.rb
```diff
@@ -51,6 +51,7 @@ def create_clock = Internal::Clock.new
         # testkit never supplies one to the MRI flavour.
         def new_instance(uri, auth_token_manager, client_certificate_manager: nil, **config)
           validate_uri(uri)
+          validate_security_settings(URI(uri).scheme, config)
```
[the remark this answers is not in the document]
klobuczek: Done — `validate_uri` now returns the parsed URI, and `new_instance` reuses it for both `validate_security_settings` and `build_connection_provider`, so the URI is parsed once instead of three times.

spec/mri/neo4j/driver/internal/driver_factory_spec.rb
klobuczek: Good call — moved to `spec/shared/neo4j/driver/security_config_spec.rb` (dropped the `Loader.mri?` gate). The behaviour holds on both flavours: JRuby's Java driver raises the same `ClientException` via SecurityPlans, with matching value-equality semantics (isCustomized). So both flavours now exercise it.

PR #463, opened 2026-09-03
fix(mri): enable tuned TCP keepalive on Bolt connections

klobuczek:
Enable `SO_KEEPALIVE` with tuned intervals (~60s idle, 10s apart, 3 probes) on the Bolt TCP socket, so a **dead peer** — a crashed host or a silent network partition that never delivers FIN/RST — is detected and the blocked read fails, instead of the socket sitting `ESTABLISHED` until the OS default idle (~2h on Linux) or, with nothing in flight, forever.

**Scope / framing (deliberately narrow):**
- General pooled-connection hygiene — not a bug fix for a specific server, and **not** a parity claim. It helps against any Bolt server, and matters most where there's no other safety net (a server that omits `connection.recv_timeout_seconds`).
- **No new public API.** Just a standard socket option, the same kind the Java driver's Netty layer uses.
- **No effect on slow-but-alive servers** — their TCP stack keeps ACKing keepalive probes, so keepalive only fires on a genuinely unreachable peer; long-running queries are unaffected. This is why it's safe where a client-side *read* timeout would not be.
- Per-interval options are platform-specific (`TCP_KEEPIDLE` on Linux, `TCP_KEEPALIVE` on macOS/BSD) and best-effort; `SO_KEEPALIVE` alone still applies with OS-default timing where they're unavailable.

MRI flavor only (the pure-Ruby Bolt stack).

### History / what was dropped
This PR originally also added an opt-in `connection_read_timeout` client-side read timeout (with a fake-server regression spec) and a connection stress/leak tool. Those were removed after review concluded:
- A client-side read timeout can't distinguish a slow query from a dead connection, so it risks killing long-running queries — which is exactly why the official driver deliberately omits it. Shipping it would mean inventing an API the reference rejects, for an edge case (hint-less server + frozen-but-alive peer) that the official driver shares and that isn't a defect against supported Neo4j (4.3+ always sends the hint).
- The leak investigation found **no leak** (steady `live_slots`/threads/connections over 40k sessions); the stress tool can return as its own PR if wanted (it's preserved in this branch's git history).

Reshaped down to the one change that stands on its own merit.

## Summary by CodeRabbit

- **Reliability**
  - Enabled TCP keepalive for network connections to help detect inactive or disrupted connections.
  - Added best-effort platform-specific keepalive tuning, with unsupported settings safely ignored.


klobuczek:
## Memory-leak investigation: not reproduced

The original report paired "hang forever" with "memory leak." This PR fixes the hang; the leak was investigated separately and **could not be reproduced under normal operation**. Recording the evidence here.

Method: a new on-demand harness (`tools/connection_stress.rb`, `rake stress`) run against Memgraph on the MRI flavor — session churn (success + failure paths), driver churn, sustained concurrency on a capped pool, and a large streamed result — sampling process RSS, thread count, live `Bolt::Connection` objects, and Ruby heap live-object count.

**20 batches × 40 threads × 50 sessions (~40k sessions):**

| Signal | Result | Interpretation |
|---|---|---|
| Ruby `live_slots` | **flat ~77,590** (spread <300 over the whole run) | no object leak — the decisive signal |
| live `Connection` objects | constant (= pool size) | no connection leak |
| threads | constant (main + one reader per pooled conn) | reader threads reaped on close |
| 300k-row stream | **+544 KB** RSS | streaming is bounded, not buffered |
| RSS | grows then **decelerates/plateaus** | malloc/allocator fragmentation, **not** a leak |

A real leak grows `live_slots`/object counts without bound; here they're flat while RSS merely plateaus (tunable via `MALLOC_ARENA_MAX`/jemalloc if desired). Reader threads are joined on `close`/`discard`, and `session#run` releases the connection on every exception path.

**Conclusion:** no evidence of a driver memory/thread/connection leak in normal operation. The most likely explanation for the original observation is the *hang-induced* thread-orphaning this PR's read-timeout fix now prevents (a hung read can no longer strand a connection + its reader). The "not tested outside rspec with a small dataset" concern is also addressed by this harness (40k sessions, real concurrency, 300k-row streaming).

klobuczek:
Addressed the review feedback in `1bca7b8`:

- **`SO_KEEPALIVE` value** (Copilot) — pass `1` instead of `true`. (Ruby's `setsockopt` does accept `true`, and the stress run opened connections fine, so it wasn't a functional bug — but `1` matches the adjacent `TCP_NODELAY, 1` and is the conventional form.)
- **Spec asserts the specific exception** (CodeRabbit) — the success path now requires `ConnectionReadTimeoutException` rather than any `Neo4jException`, so an unrelated connection failure can't pass as success. The stalled/hung/other-error outcomes each fail with a clear message.
- **Join the fake-server thread** (Copilot) — `thread.kill.join` in the ensure, so no background thread leaks between examples.

Spec still green (raises `ConnectionReadTimeoutException` in ~2s); rubocop clean on the changed files.

klobuczek:
Reshaped (force-push `d2ba47b`) to **keepalive-with-tuned-intervals only**. Dropped the `connection_read_timeout` knob + its fake-server spec (a client read timeout can't tell a slow query from a dead connection — the reason the official driver omits it) and the stress tool (leak investigation found nothing; recoverable from this branch's history for a separate PR). See the updated description for the narrowed scope. Earlier comments about the read-timeout fix / leak report no longer apply to the shipped change.

